Issue #3, opened 2016-12-28
Create SteamScript


plata:
Create SteamScript to install games from Steam.

PR #5, opened 2016-12-30
added SteamScript to install steam applications

plata:
fixes: Create SteamScript #3


plata:
@qparis could you have a look if this is ok for you (at least from the interface)? I added an example application here in #6.

qparis:
I will try it tonight

qparis:
Thanks a lot

Issue #4, opened 2016-12-30
Wine.getProgramFiles() returns string with line break at the end


plata:

qparis:
Hi ! Do you want to fix it? :)

plata:
I can. But now I would like to have LocalApplicationsSource > GitApplicationsSource. Would be much easier to test changes in the scripts.

qparis:
Why don't you override the system side settings?

plata:
How?

qparis:
Oh I see, ConfigApplicationSource needs to be implemented

qparis:
(There is a ticket opened for that)

plata:
I've seen that ticket. I mean I can easily change the order in the code. But I really think it would be better the other way around by default.

qparis:
By default, it should be Local User Side > Internet > Local application side

qparis:
This is why the ticket should be implemented to set this properly :)

plata:
For me Local User Side is `${application.user.repository}`. So in fact we do not have Local application side currently and the Local User Side < Internet.

qparis:
Yep, you are correct. This is why the two tickets need to be done :)

PR #8, opened 2016-12-30
fixed: Wine.getProgramFiles()

plata:
fixes: Wine.getProgramFiles() returns string with line break at the end #4


qparis:
Awesome, thanks a lot

PR #28, opened 2016-12-31
Niko: Through The Dream

plata:


plata:
still requires:
- #26 
- #27

Games/Niko: Through The Dream/Steam/script.json
```diff
@@ -0,0 +1,7 @@
+{
+	"scriptName": "Steam",
+	"compatibleOperatingSystems": ["MACOSX", "LINUX"],
+	"testingOperatingSystems": ["LINUX"],
```
qparis: Testing means that the script is beta

plata: So it should only be set when the script is working already? (Will add a remark to the wiki to clarify).

qparis: It should be set if you are unsure that the script will run properly for everybody (like if it requires a huge amount of hack, or if it is not compatible with a any graphics cards, etc...)

plata: I see.

plata: What is requiresNoCD? (see [script.json](https://github.com/PlayOnLinux/Scripts/wiki/script.json))

qparis: It means that your program cannot work without patching the original protection. It may be illegal to install it depending on your country

plata: Updated wiki.

qparis:
Do you want to wait for dotnet40 before merging?

plata:
Yes. It does not work yet.

I just wanted to prepare my scripts to figure out what is still needed for them to work.

PR #40, opened 2017-01-01
Steam install progress bar

plata:


Functions/QuickScript/SteamScript/script.js
```diff
@@ -20,10 +21,59 @@ SteamScript.prototype.constructor = SteamScript;
...
     this._applicationHomepage = "http://store.steampowered.com/app/" + appId;
-    this._executableArgs = "steam://rungameid/" + this._appId;
+    this._executableArgs = ["-silent", "steam://rungameid/" + this._appId];
     return this;
 };
 
+SteamScript.prototype.getBytesToDownload = function(wine) {
+    // wait until download started
+    while (!fileExists(wine.prefixDirectory + "/drive_c/" + wine.getProgramFiles() + "/Steam/steamapps/appmanifest_" + this._appId + ".acf"))
+    {
```
qparis: I think we should at least sleep here (or better, implement a file watcher in Java)

Functions/QuickScript/SteamScript/script.js
```diff
@@ -20,10 +21,59 @@ SteamScript.prototype.constructor = SteamScript;
...
+    }
+
+    var processBuilder = new java.lang.ProcessBuilder(["bash", "-c", "cat appmanifest_" + this._appId + ".acf | grep BytesToDownload | egrep -o '[0-9]+'"]);
```
qparis: This could be parsed in JavaScript directly. For a start, you can use "cat" function

Functions/QuickScript/SteamScript/script.js
```diff
@@ -20,10 +21,59 @@ SteamScript.prototype.constructor = SteamScript;
...
+
+SteamScript.prototype.getBytesDownloaded = function(wine) {
+    var downloadFolder = wine.prefixDirectory + "/drive_c/" + wine.getProgramFiles() + "/Steam/steamapps/downloading/" + this._appId;
+    // download folder is not yet/no more available
+    if (!fileExists(downloadFolder))
+    {
+        // check if download already finished (download folder has been deleted)
+        if (fileExists(wine.prefixDirectory + "/drive_c/" + wine.getProgramFiles() + "/Steam/steamapps/appmanifest_" + this._appId + ".acf"))
+        {
+            var processBuilder = new java.lang.ProcessBuilder(["bash", "-c", "cat appmanifest_" + this._appId + ".acf | grep BytesDownloaded | egrep -o '[0-9]+'"]);
```
qparis: Same here

Functions/QuickScript/SteamScript/script.js
```diff
@@ -20,10 +21,59 @@ SteamScript.prototype.constructor = SteamScript;
...
+            var process = processBuilder.start();
+            return Number(org.apache.commons.io.IOUtils.toString(process.getInputStream()).trim());
+        }
+        else
+        {
+            return 0;
+        }
+    }
+
+    var processBuilder = new java.lang.ProcessBuilder(["bash", "-c", "du -sb *_Data | egrep -o '[0-9]+'"]);
```
qparis: I think we have some beans to get directory size

Functions/QuickScript/SteamScript/script.js
plata: How can I sleep? I wanted to do that originally but all I couldn't really find something on the Internet.

qparis: java.lang.Thread.sleep() but it would be better to create a bean

Functions/QuickScript/SteamScript/script.js
qparis: I was more thinking about Files.size(File directory)

DirectoryWatcherSize needs to needs to be updated but the way it works is that it opens a new Thread, regularly checks a directory file size and accepts a callback you have defined.

It is definitely a good long term solution but AbstractDirectoryWatcher seems to be broken. We also need to create the corresponding factories

Functions/QuickScript/SteamScript/script.js
qparis: You can also create a JavaScript Function Script (Threading) and define a sleep function inside

Functions/QuickScript/SteamScript/script.js
plata: Will have to add `size()` to FileUtilities then.

qparis: Or you can directly define "du" function in Functions, Filesystem, Files which will directly use java.nio.file.Files

qparis:
All right, appmanifest looks like really similar to json.
Maybe we should implement a full parser later

Functions/QuickScript/SteamScript/script.js
plata: Wherever I look, they say that you cannot easily retrieve the size of a directory without iterating over its contents.

qparis: Well yes this is correct, but we should at least avoid to spawn a process just for this

plata:
> All right, appmanifest looks like really similar to json.

In fact, [SteamShutdown](https://github.com/akorb/SteamShutdown) converts the file to json first.

qparis:
How do they manage to do this?

plata:
Can't tell right away. Would have to look into the code as well. Just remember that I saw some json there.

Functions/QuickScript/SteamScript/script.js
plata: I'm trying to enhance the FileUtilities now. But I have a problem with the repository config. I tried `file+${application.user.repository}` but that results in 
```
Bean instantiation via factory method failed
```

qparis: Try file://${} instead

plata: Can I somehow set this review to resolved?

PR #66, opened 2017-01-03
Allow to set OS per application

plata:
fixes: #65

### DIFF
--- a/Functions/Engines/Wine/script.js
+++ b/Functions/Engines/Wine/script.js
@@ -425,3 +425,32 @@ Wine.prototype.overrideDLL = function() {
     return new OverrideDLL()
         .wine(this)
 };
+
+var SetOsForApplication = function() {
+    var that = this;
+    that._regeditFileContent =
+        "REGEDIT4\n" +
+        "\n"+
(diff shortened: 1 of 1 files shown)


plata:
I followed the example of `OverrideDLL()`. However, I don't quite understand why it is done this way. To use it, you have to write:
```javascript
wine.setOsForApplication().set("Notepad++.exe", "win7").do();
```
Why can't it be a simple function which could be used like:
```javascript
wine.setOsForApplication("Notepad++.exe", "win7");
```

plata:
I think I understand now. It's because it allows you to set multiple overrides at once, right?

qparis:
Yeah exactly ! Don't forget that we now have beans to parse and generate .reg files

qparis:
I'll review your code when I am home

plata:
Didn't know about the beans. In this case, probably the code for OS per application, sound driver and DLL override should be rewritten. I guess it would be the best to do everything in RegistryWinePrefixConfiguration.java directly so it can be used also from other UIs.

Can we merge OS per application and sound driver as they are and create an issue to do that in PoL 5?

PR #69, opened 2017-01-04
Wine refinements

plata:


Functions/Engines/Wine/script.js
```diff
@@ -8,333 +8,392 @@ LATEST_STABLE_VERSION = "1.8.6";
...
+/**
+*
+* @param {string} [debug]
+* @returns {string|Wine}
+*/
+Wine.prototype.debug = function (debug) {
+    // get
+    if (arguments.length == 0) {
+        this._wineDebug = "";
+        return this._wineDebug;
```
qparis: This will always return an empty string. What if I want to get the current value?

plata: see #82

PR #70, opened 2017-01-04
Ether One Redux

plata:


Games/Ether One Redux/Steam/script.js
```diff
@@ -0,0 +1,14 @@
+include(["Functions", "QuickScript", "SteamScript"]);
+
+new SteamScript()
+    .name("Ether One Redux")
+    .editor("White Paper Games")
+    .author("Plata")
+    .appId(391920)
+    .wineVersion("1.9.23")
+    .wineArchitecture("amd64")
```
qparis: Is it required?

plata: The architecture? Yes, I think so.

PR #74, opened 2017-01-04
implemented wine.managed()

plata:
fixes: Implement wine.setManaged() #26


qparis:
Conflicts need to be resolved

PR #81, opened 2017-01-04
Quantum Conundrum

plata:


Games/Quantum Conundrum/Steam/script.js
```diff
@@ -0,0 +1,12 @@
+include(["Functions", "QuickScript", "SteamScript"]);
+include(["Functions", "Verbs", "msvc90"]);
+
+new SteamScript()
+    .name("Quantum Conundrum")
+    .editor("Square Enix")
+    .author("Plata")
+    .appId(200010)
+    .preInstall(function(wine, wizard) {
+        wine.msvc90();
```
qparis: Does it work if you simply install vcrun2008?

plata: As I said in the msvc90 discussion, it does not (tried that first). No idea why, though.

Issue #39, opened 2017-01-01
Implement wine.useNativeExtension()


plata:
e.g. for PDF

PR #94, opened 2017-01-06
use native application for certain file extension

plata:
fixes: Implement wine.useNativeExtension() #39


plata:
Everything look ok, but I cannot find the path with `regedit`.

PR #96, opened 2017-01-06
install gecko and mono

plata:
fixes: [PoL issue #396](https://github.com/PlayOnLinux/POL-POM-5/issues/396)

### DIFF
--- a/Functions/Engines/Wine/script.js
+++ b/Functions/Engines/Wine/script.js
@@ -329,7 +329,9 @@ Wine.prototype._installVersion = function () {
             if (distribution.name == fullDistributionName) {
                 distribution.packages.forEach(function (winePackage) {
                     if (winePackage.version == version) {
-                        that._installWinePackage(wizard, winePackage, localDirectory)
+                        that._installWinePackage(wizard, winePackage, localDirectory);
+                        that._installGecko(wizard, winePackage, localDirectory);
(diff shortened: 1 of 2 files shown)


Functions/Engines/Wine/script.js
```diff
@@ -358,6 +360,38 @@ Wine.prototype._installWinePackage = function (setupWizard, winePackage, localDi
...
+        .checksum(winePackage.monoMd5)
+        .algorithm("md5")
+        .name(winePackage.monoFile)
+        .directory("mono")
+        .get();
+
+    var wineMonoDir = localDirectory + "/share/wine/mono";
+    mkdir(wineMonoDir);
+
+    lns(mono, wineMonoDir + "/" + winePackage.monoFile);
```
qparis: The whole mono directory should be symlinked, same for gecko

plata: Ok... But then there will be all gecko/mono versions, not just the ones belonging to this wine?

qparis: Yes so if you install any gecko package, it will update all wine versions. It will allow to fix the urls if required (and it is often required because we can be late sometimes)

plata: Fine. Will be done in a sec.

PR #105, opened 2017-01-08
Fix SteamScript long text not showed correctly

Brainzyy:
Fixes https://github.com/PlayOnLinux/POL-POM-5/issues/449


plata:
This should be fixed in PoL directly (wrap the text).

Brainzyy:
@plata Alright, then I'll close this.

PR #112, opened 2017-01-08
fixed script.json for Hexcells

plata:


qparis:
Is this game free? It could be tested on OSX

plata:
No. But cheap.

qparis:
I'm leaving it now but we need to find a way to efficently test apps on all OSes

PR #113, opened 2017-01-08
STAR WARS Jedi Knight games 

plata:
I'm having troubles testing them but wanted to share in case someone wants to give it a try.


plata:
We will probably have similar situations (i.e. unfinished/not fully tested scripts) rather often in the future. Therefore I propose to add a staging/testing branch where such scripts can be committed. Once they are finished/tested, they will go to master.

qparis:
There is a testing status for that. We can also create a third git repository

plata:
I know about the testing status. However, I assume that people will not really look at the testing status but rather assume that scripts in master are working (which is reasonable in my opinion). 
Why do you prefer a separate repository to a branch? A branch would be much easier to maintain, wouldn't it?

qparis:
Yes but we need to allow the client to create a per branch application source in this case

plata:
If we want to show those scripts in PoL by default, yes. Not so sure if we should do that, though. I would see it more as "alpha" state scripts which should not show up in PoL. If someone wants to test them, she should checkout from GitHub manually.

qparis:
That make sense. We should also create some integration test for the scripts by the way

plata:
Yes. Do you have any idea how this could be done?

What would you prefer: testing or staging? I would create the branch and update the wiki accordingly.

qparis:
Testing is the branch included in POL, so I would call it staging to avoid confusion

qparis:
Most of these game work on OSX. Anyway I have them and I'll get them tested. Can you change the OS? :)

plata:
I thought again about the staging branch. It will be quite difficult to pull only a certain script to master with one staging branch. Therefore I think the better solution would be to simply leave the pull request open until it's finished. To avoid loosing track of the pull requests which are finished, we can use labels.
The only problem that might occur with this solution is if the one who created the pull request does not care anymore. However, I believe that this will be a rare case. If it happens, either someone else can fork and create a new pull request or we decide to take over and create a branch for it.

PR #114, opened 2017-01-09
fixed wine architecture

plata:
Use default value only if nothing else set.

fixes: PlayOnLinux/POL-POM-5#463

### DIFF
--- a/Functions/Engines/Wine/script.js
+++ b/Functions/Engines/Wine/script.js
@@ -122,8 +122,7 @@ Wine.prototype.prefix = function (prefix) {
 
     this._prefixConfiguration.writeValue("wineDistribution", this._distribution);
 
-    var defaultArchitecture = Bean("architectureFetcher").fetchCurrentArchitecture().getNameForWinePackages();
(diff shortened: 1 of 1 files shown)


Functions/Engines/Wine/script.js
```diff
@@ -122,10 +122,8 @@ Wine.prototype.prefix = function (prefix) {
 
     this._prefixConfiguration.writeValue("wineDistribution", this._distribution);
 
-    if (!this._architecture) {
-        var defaultArchitecture = Bean("architectureFetcher").fetchCurrentArchitecture().getNameForWinePackages();
-        this._architecture = this._prefixConfiguration.readValue("wineArchitecture", defaultArchitecture);
-    }
+    var defaultArchitecture = Bean("architectureFetcher").fetchCurrentArchitecture().getNameForWinePackages();
```
qparis: I thought we wanted x86 as a default architecture?

plata: You're completely right.

Issue #118, opened 2017-01-10
implement quicktime76


Brainzyy:
Needed for games like Audiosurf.

Brainzyy:
@plata it seems like this was not implemented correctly. Audiosurf can't find Quicktime.

plata:
I couldn't test it. Can you compare the implementation to the one of winetricks?

Brainzyy:
@plata I think I know what was wrong. Quicktime gets installed silently here, no? That doesn't work with Quicktime.

plata:
Yes it is silently (`/qn`). I just followed the winetricks way of doing it. Does it work if you remove the `/qn`?

Brainzyy:
@plata That fixed it :D going to prep a PR

PR #122, opened 2017-01-14
quicktime76

plata:
fixes: implement quicktime76 #118


plata:
@Brainzyy can you try if it works?

Functions/Verbs/quicktime76/script.js
```diff
@@ -0,0 +1,17 @@
+include(["Functions", "Engines", "Wine"]);
+include(["Functions", "Net", "Resource"]);
+include(["Functions", "Verbs", "luna"]);
```
qparis: This is not required

plata: True.

PR #125, opened 2017-01-14
Assassin’s Creed II

plata:


plata:
@qparis how did you get Uplay working for AC: Brotherhood?

Games/Assassin’s Creed II/Steam/script.json
```diff
@@ -0,0 +1,7 @@
+{
+	"scriptName": "Steam",
+	"compatibleOperatingSystems": ["LINUX"],
+	"testingOperatingSystems": ["MACOSX"],
```
qparis: Cannot be testing if it is not compatible :)

qparis:
Should work as it should'nt it?

plata:
I thought so as well. But Uplay keeps crashing at start. So I thought you should have the same problem with AC: Brotherhood.

Brainzyy:
@plata @qparis Just a heads up, I heard that a recent UPlay update broke compatibility with wine.

plata:
@Brainzyy thanks for the remark. I'm trying now to write a Uplay script nevertheless.

Brainzyy:
@plata You're welcome, you can still login to UPlay with a prefix set to Windows Vista and latest Wine Staging, but according to that bug, games don't work anymore.

qparis:
Agreed, we should remove AC:B for the moment

plata:
@qparis just set it to testing. Would be a shame if we would lose the script completely.

qparis:
Maybe we should also create a UPlaySteamScript template

plata:
Yes. Also have been thinking about this. Just wait a sec I will push my UPlay script.

Brainzyy:
@plata @qparis A seperate UplayScript would be useful as well. Since some games are not on steam, but only on Uplay.

plata:
@Brainzyy I think it is not possible to start a Uplay installation from a script.

plata:
Uplay does not crash anymore. However, it's still not possible to play the game. Uplay shows "There's a problem connecting to the internet..." and "Unable to download Uplay game files". 
Maybe this is related to
```
fixme:file:MoveFileWithProgressW MOVEFILE_WRITE_THROUGH unimplemented
```
or
```
fixme:crypt:CRYPT_VerifyChainRevocation Unable to find CRL for CA certificate
```

plata:
The script itself is fine and can be merged. The problem is #221.

PR #126, opened 2017-01-14
Uplay

plata:


plata:
@Brainzyy it has problems even with staging and vista.

Brainzyy:
@plata Hmm, what kind of problems?

plata:
I think it's something about xaudio2_7.
```
err:ole:COMPOBJ_DllList_Add couldn't load in-process dll L"C:\\windows\\system32\\xaudio2_7.dll"
err:ole:CoGetClassObject no class object {5a508685-a254-4fba-9b82-9a24b00306af} could be created for context 0x1
```

Brainzyy:
@plata That is normally fixed by installing xact_jun2010.

Brainzyy:
@plata I just tried with plane Wine and I can open and login to Uplay just fine with staging 2.0-rc4 set to Vista.

plata:
@Brainzyy can you checkout this pull request (or the branch on my fork) and try that?

Brainzyy:
@plata I can't seem to make it work.

plata:
@Brainzyy could it be that 2.0rc-3 is not enough? I couldn't install 2.0rc-4 with PoL.

Brainzyy:
@plata Since Wine is in code freeze, there shouldn't be any commits between 1.9.24 (which I confirmed working) and rc3 and rc4. So that should not be the issue.

plata:
Hm. That's strange. There must be a difference between plain wine and PoL somehow.

@qparis any idea?

qparis:
I think Uplay should be treated like Steam, we should not have a single prefix for all UPlay games

plata:
@qparis that's definitely true. Nevertheless I think it makes sense to have a Uplay script because we also have a Steam script. 

The question here is why Uplay works with plain wine but not with PoL.

qparis:
Have you tried exactly the same versions?

plata:
@Brainzyy did the comparison. I tried several wine versions with PoL (also the ones which seem to work with plain wine, e.g. 2.0-rc1 staging as on AppDB) .

qparis:
Any error output on your terminal that you do not have with wine?

plata:
Well, I got the xaudio stuff I described above. Maybe check out this PR and try yourself.

qparis:
https://www.playonlinux.com/wine/logs/linux-x86/wine-2.0-rc5-linux-x86.log

plata:
What's that?

qparis:
Our build log. I'm trying to install other dependencies

plata:
Ok sorry for being unclear: I saw it's the build log but I'm unsure what you want to show me.

qparis:
We have three dependencies missing:
configure: OpenCL 32-bit development files not found, OpenCL won't be supported.
configure: libhal 32-bit development files not found, no legacy dynamic device support.
configure: gstreamer-1.0 base plugins 32-bit development files not found, gstreamer support disabled
configure: libudev 32-bit development files not found, plug and play won't be supported.

plata:
Why don't we get an error then?

qparis:
There are not required dependencies. I'm trying to rebuild wine rc5 considering this. Stay tuned

Brainzyy:
@qparis @plata Ah yes, I hadn't thought of that. For instance, I usually build my engines with msxml2, while others don't and then just install msxml3 through winetricks. To get to my point, I can install .net4 for instance without winetricks, while without the msxml2 being compiled into my wine, I have to winetrick msxml3 to be able to install it.

plata:
@Brainzyy regarding .NET: I think there is a problems also with mono (see #130).

qparis:
The problem with mono won't affect .NET ; it will just ask the user to download it

Brainzyy:
@plata Is mono even useful? I always just install the correct .Net version through wineskin. I've never had a use for mono. 

Btw here al my build options @qparis: `--disable-option-checking --without-oss --without-cms --disable-tests --without-v4l --without-alsa --without-audioio --without-capi --with-coreaudio --without-esd --without-hal --without-jack --with-xcomposite --with-xcursor --with-xinerama --with-xinput --with-xml --with-xrandr --with-xrender --with-xshape --with-xshm --with-xslt --with-xxf86vm --with-x --without-fontconfig --without-gphoto --without-gstreamer --without-dbus`

qparis:
It seems that you have less stuff than POL's version of wine
Mono is useful yes, we want to avoid the download prompt

qparis:
Can you retry with the last rc5?

plata:
There's no 2.0-rc5 staging. I can try upstream but AppDB says you need staging.

qparis:
My bad

plata:
2.0-rc5 upstream doesn't work.

Brainzyy:
@plata Yeah, it really needs staging. Since Wine is in code freeze at the moment, it's going to take a while before the, modified or original, staging fixes get committed upstream.

qparis:
Staging rc4 should be available in POL

plata:
Just tried again with rc4 staging. Uplay is still crashing.

PR #128, opened 2017-01-14
Make quicktime not silent to fix it (Fixes #118)

### LINKED ISSUE #118: implement quicktime76

Brainzyy:
fixes #118  
Thanks @plata for implementation

### DIFF
--- a/Functions/Verbs/quicktime76/script.js
+++ b/Functions/Verbs/quicktime76/script.js
@@ -6,7 +6,7 @@ Wine.prototype.quicktime76 = function() {
         .name("QuickTimeInstaller.exe")
         .get();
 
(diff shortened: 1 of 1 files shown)


plata:
The only thing which is still strange to me is that silent seems to work with winetricks.

Brainzyy:
@plata For me it isn't silent in winetricks

Brainzyy:
@plata Just checked out the winetricks source, and as far as I can see it is not silent by default. I think you need to specify the winetricks flag -q which has the behavior described above.

plata:
Ok. So it's also not working with `winetricks -q quicktime76`. It's fine. Just wanted to understand what's going on. Do the other parameters which are passed to the installation make any sense if it is not silent?

Brainzyy:
@plata it does work with -q, but it isn't actually silent. It just uses autohotkey to do the navigation through the installation windows really fast. And I'll check out the other arguments, but it's fine for now.

PR #134, opened 2017-01-15
Tropico 3

plata:


Games/Tropico 3/Steam/script.js
```diff
@@ -0,0 +1,9 @@
+include(["Functions", "QuickScript", "SteamScript"]);
+
+new SteamScript()
+    .name("Tropico 3")
+    .editor("Haemimont Games")
+    .author("Plata")
+    .appId(23490)
+    .wineVersion("2.0-rc5")
```
qparis: Is it required?

plata: The 2.0-rc5? I don't think so. But I thought I'd take that, then we can go to stable once 2.0 is released.

PR #135, opened 2017-01-15
Age of Empires II HD

plata:


qparis:
Have you tested it? We have a bug report on POL4

plata:
Just played it for quite a while. The launcher is broken but apart from that it's fine.

PR #143, opened 2017-01-27
STAR WARS Battlefront II: use wine stable

plata:


plata:
If I set the resolution > 1920*1080, the game crashes. Do we have place to store such information?

qparis:
In the description?

plata:
Don't know. For me, the description is about the game itself (like it's a strategy game etc.)

PR #147, opened 2017-01-28
Ether One Redux: use wine stable

plata:


plata:
The Steam game overlay needs to be disabled. Otherwise you get texture bugs.

@qparis can this be done in PoL? If not, how do we communicate this?

qparis:
In POLv5, we could do 
 POL_Wine_OverrideDLL "" "gameoverlayrenderer"

plata:
Can we add that functionality to SteamScript as `gameOverlay()`? 
To be used like:
```javascript
SteamScript.gameOverlay(false)
```

qparis:
sure

PR #156, opened 2017-01-29
custom installer: let user select wine settings

plata:
Would be great to have a bean to list the available options in a combobox.

### DIFF
--- a/Functions/QuickScript/InstallerScript/script.js
+++ b/Functions/QuickScript/InstallerScript/script.js
@@ -28,6 +28,13 @@ InstallerScript.prototype.go = function() {
     // get installation file from concrete InstallerScript implementation
     var installationFile = this._installationFile(setupWizard);
 
+    // let user select wine settings if desired
+    if (this._wineUserSettings) {
+        this._wineArchitecture = setupWizard.textbox("Please enter the wine architecture.", this._wineArchitecture);
+        this._wineDistribution = setupWizard.textbox("Please enter the wine distribution.", this._wineDistribution);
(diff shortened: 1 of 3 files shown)


Functions/QuickScript/InstallerScript/script.js
```diff
@@ -28,6 +28,13 @@ InstallerScript.prototype.go = function() {
     // get installation file from concrete InstallerScript implementation
     var installationFile = this._installationFile(setupWizard);
 
+    // let user select wine settings if desired
+    if (this._wineUserSettings) {
+        this._wineArchitecture = setupWizard.textbox("Please enter the wine architecture.", this._wineArchitecture);
+        this._wineDistribution = setupWizard.textbox("Please enter the wine distribution.", this._wineDistribution);
```
qparis: These should not be a textbox

plata: Yes. Have a look at my comment in the PR.

qparis:
We already have a lib to list available wine versions

Functions/QuickScript/InstallerScript/script.js
qparis: We already have a lib to list available wine versions :)

plata: See it now...

qparis: We have Wine.prototype.getAvailableVersions

plata: Yes I found that as well now :)

PR #163, opened 2017-02-06
allow to set Windows version in custom installer

plata:
fixes PlayOnLinux/POL-POM-5#540

### DIFF
--- a/Custom/LocalInstaller/Local/script.js
+++ b/Custom/LocalInstaller/Local/script.js
@@ -6,4 +6,8 @@ new LocalInstallerScript()
     .author("Plata")
     .category("Custom")
     .wineUserSettings(true)
+    .preInstall(function(wine, wizard) {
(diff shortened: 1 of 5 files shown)


qparis:
I'm really not convinced that quick script should have all wine properties. preInstall and' postInstall is the right thing to handle this

plata:
Don't think that's possible. Wine needs to know architecture, version etc when you call `prefix()`.

plata:
Could do the Windows version in the `preInstall` but would be strange to handle this one setting differently than the others.

qparis:
Which others?

plata:
architecture, version, distribution

qparis:
Those are POL settings, this is totally different.
You have hundred of wine settings (windows version, sound driver, dll overrides, glsl, etc...). It is just impossible to have setters for each of them

plata:
I put it to preInstall.

PR #164, opened 2017-02-11
set default values in custom installers

plata:
Helps the user in deciding which wine architecture, distribution, version and Windows version shall be used.

requires PlayOnLinux/POL-POM-5#584


qparis:
This needs to be fixed :)

plata:
Fixed.

PR #174, opened 2017-02-26
moved apps into separate "Apps" directory

plata:
Allows for a clearer destinction between apps and functions in the directory structure.

requires PlayOnLinux/POL-POM-5#622


madoar:
@plata is this PR still open?

plata:
Not really. We have to decide PlayOnLinux/POL-POM-5#745 first.

PR #175, opened 2017-02-27
implemented generic install/delete for Wine

plata:


Functions/Engines/Wine/script.js
```diff
@@ -287,6 +288,47 @@ Wine.prototype.getAvailableVersions = function () {
 };
 
 /**
+* install
+* @param {string} category
+* @param {string} subCategory
+* @param {string} version
+* @param {json} userData
+*/
+Wine.prototype.install = function (category, subCategory, version, userData) {
```
plata: If I'm not mistaken there was not even a public interface to install a Wine version at all.

PR #177, opened 2017-03-01
update LoL Beta script

feanor12:
This script should work.
It creates a shortcut for the launcher and one for the LeagueClient.exe.

### DIFF
--- a/Games/League of Legends/BETA Client/script.js
+++ b/Games/League of Legends/BETA Client/script.js
@@ -37,9 +37,8 @@ new CustomInstallerScript()
         //Create the installation directory
         mkdir(wine.prefixDirectory+"drive_c/LoL");
         //Create the early LeagueClient.exe for shortcut-creation
-        var launcher = wine.prefixDirectory+"drive_c/LoL/lol.launcher.admin.exe"
-        var client = wine.prefixDirectory+"drive_c/LoL/LeagueClient.exe"
-        lns(launcher,client)  
+        var client = wine.prefixDirectory+"drive_c/LoL/LeagueClient.exe";
+        writeToFile(client,"");  
         //Create the shortcut for the beta client
         new WineShortcut()
             .name("BETA Client")
@@ -54,9 +53,9 @@ new CustomInstallerScript()
     })
     .postInstall(function(wine,wizard) {
         //Remove the temporary directory
-        remove(wine.prefixDirectory+"drive_c/LoL_tmp")
+        remove(wine.prefixDirectory+"drive_c/LoL_tmp");
         //Remove the fake BETA Client, because it wont update otherwise
-        remove(wine.prefixDirectory+"drive_c/LoL/LeagueClient.exe")
+        remove(wine.prefixDirectory+"drive_c/LoL/LeagueClient.exe");
     })
     .executable("lol.launcher.admin.exe")
     .go();
--- a/Games/League of Legends/BETA Client/script.js
+++ b/Games/League of Legends/BETA Client/script.js
@@ -4,11 +4,20 @@ include(["Functions", "Verbs", "sandbox"]);
 include(["Functions", "Verbs", "d3dx9"]);
 include(["Functions", "Verbs", "vcrun2015"]);
 
+// Installs League of Legends and creates two shortcuts.
+//
+// One shortcut called League of Legends which starts the launcher
+// and one shortcut called BETA Client which starts the BETA client
+//
+// Important: 
+// The BETA-Client has to be installed using the launcher.
+// After the upgrade the BETA Client will do the game patching.
+
 new CustomInstallerScript()
     .name("League of Legends BETA Client")
     .editor("Riot Games")
     .applicationHomepage("http://euw.leagueoflegends.com/")
-    .author("Plata,feanor12")
+    .author("Plata, feanor12")
     .installationCommand(function(wizard) {
         return {command: "msiexec", args: ["/i", "C://LoL_tmp/Lol.EUW.msi","APPDIR=C:\\LoL","/q"]};
     })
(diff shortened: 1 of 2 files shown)


plata:
Could you add some comments in the script (e.g. to explain the hack for the shortcut)?

Games/League of Legends/BETA Client/script.js
```diff
@@ -1,32 +1,47 @@
...
             .name("LeagueofLegends_EUW_Installer_2016_11_10.exe")
             .get();
 
+        mkdir(wine.prefixDirectory+"drive_c/LoL_tmp");
         mkdir(wine.prefixDirectory+"drive_c/LoL");
-        wine.run(setupFile,["/extract:C:\\LoL","/exenoui"]).wait();
+        var launcher = wine.prefixDirectory+"drive_c/LoL/lol.launcher.admin.exe"
+        var client = wine.prefixDirectory+"drive_c/LoL/LeagueClient.exe"
+        lns(launcher,client)           
+        new WineShortcut()
```
plata: Do you need this additional shortcut? If the launcher overwrites LeagueClient.exe, it should be enough to have a shortcut for LeagueClient.exe (as it links to the launcher originally).

feanor12: The problem is that the launcher wont update to the beta client if the LeagueClient.exe exists.
Is there a function like touch to create the file?

plata: You could use writeToFile. But if the launcher does not overwrite an existing LeagueClient.exe, it will not work anyways. I leave it for @qparis to also have a look (maybe he has another idea). Apart from that I think we can merge this.

feanor12: Ok I'll try using writeToFile.

Games/League of Legends/BETA Client/script.js
```diff
@@ -1,32 +1,61 @@
...
-
+        //Download the setup file
         var setupFile = new Resource()
             .wizard(wizard)
             .url("https://riotgamespatcher-a.akamaihd.net/ShellInstaller/EUW/LeagueofLegends_EUW_Installer_2016_11_10.exe")
             .checksum("cbbf9883a59d2dada87e6aa8cc4484b07ea797da")
             .name("LeagueofLegends_EUW_Installer_2016_11_10.exe")
             .get();
-
+        //Create the temporary directory
```
qparis: In general, this sort of comment is not required: a good code should speak by itself. 

A small tip that can help you: 
Ask yourself if you are describing a general behavior or if you are explaining your code. 
* In the first case, as pointed out by @plata, you should comment your code.
* In the second case, either the comment is redundant, either your code is not clear enough to be directly understandable. If you want to make your code better, a very good method is to create functions with a name explaining what t does instead of writing comments

Games/League of Legends/BETA Client/script.js
```diff
@@ -1,32 +1,61 @@
...
+
 new CustomInstallerScript()
     .name("League of Legends BETA Client")
     .editor("Riot Games")
     .applicationHomepage("http://euw.leagueoflegends.com/")
-    .author("Plata")
+    .author("Plata, feanor12")
     .installationCommand(function(wizard) {
-        return {command: "msiexec", args: ["/i", "C://LoL/Lol.EUW.msi","APPDIR=C:\\LoL","/q"]};
+        return {command: "msiexec", args: ["/i", "C://LoL_tmp/Lol.EUW.msi","APPDIR=C:\\LoL","/q"]};
```
qparis: I think that we should update wine runners so that it can directly run msi files

Games/League of Legends/BETA Client/script.js
```diff
@@ -1,32 +1,61 @@
...
+//
+// One shortcut called League of Legends which starts the launcher
+// and one shortcut called BETA Client which starts the BETA client
+//
+// Important: 
+// The BETA-Client has to be installed using the launcher.
+// After the upgrade the BETA Client will do the game patching.
+
 new CustomInstallerScript()
     .name("League of Legends BETA Client")
```
madoar: What's missing here is the information that this script installs the EUW client. 

I believe that we want to support the other region clients too.
To do this, I don't think that we want to add an install script for every region. 
I believe it is better to ask the user during the installation, which region client he wants to install. 
This can be done the same way as with the language selection in the [Internet Explorer 7](https://github.com/PlayOnLinux/Scripts/blob/master/Internet/Internet%20Explorer%207.0/Online/script.js) script.

plata: Yes. You can do this in the preInstall.

Games/League of Legends/BETA Client/script.js
plata: How do you know the msiexec parameter (`/i`, `/a`...) then? (I don't think that you want to introduce a new function for msi, do you?)

Games/League of Legends/BETA Client/script.js
feanor12: Now you can do the install for different regions.

Games/League of Legends/BETA Client/script.js
```diff
@@ -12,50 +12,123 @@ include(["Functions", "Verbs", "vcrun2015"]);
...
+                       "Latin America North",
+                       "Latin America South",
+                       "Oceania",
+                       "Japan",
+                       "Turkey",
+                       "Brasil",
+                       "EU Nordic & East",
+                       "North America",
+                       "Russia"];
+        var selectedRegion = wizard.menu("Select your region:", regions);
```
plata: If you want to give a default region, have a look at the [wiki](https://github.com/PlayOnLinux/Scripts/wiki/SetupWizard#query-user-input)

Games/League of Legends/BETA Client/script.js
```diff
@@ -12,50 +12,123 @@ include(["Functions", "Verbs", "vcrun2015"]);
...
+        var reguionID, url, sha1;
+        var baseUrl = "https://riotgamespatcher-a.akamaihd.net/ShellInstaller/";
+        switch(selectedRegion.text){
+            case "EU West":
+                reguionID = "EUW";
```
plata: probably: regionID. However, regionID is never used. I guess you wanted to do something like:
```Javascript
url = baseUrl+regionID+"/LeagueofLegends_"+regionID+"_Installer_2016_05_26.exe";
```
But then you noticed that the dates are different, so you cannot use it?

Games/League of Legends/BETA Client/script.js
feanor12: Can I disable the empty selection or do I have to create a loop and check for it?

Games/League of Legends/BETA Client/script.js
plata: Ah ok. Didn't see that. Then just fix the typo.

Games/League of Legends/BETA Client/script.js
plata: If you cannot use it with the wrong region, you're right that setting a default doesn't make sense.

The empty selection should be disabled in StepRepresentationMenu.java. For StepRepresentationTextBox, this has been implemented already:
```Java
if (newValue.trim().equals("")) {
    setNextButtonEnabled(false);
}
```
For the menu it should be:
```Java
listViewWidget.getSelectionModel().selectedItemProperty().addListener
```

feanor12: OK, I added
```
 listViewWidget.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> {
            if (newValue.trim().equals("")) {
                setNextButtonEnabled(false);
            } else {
                setNextButtonEnabled(true);
            }
        });
```
How do I run PoL in eclipse?

plata: Don't know about eclipse. But JavaFXApplication should be the correct target. Can we take this as a new issue to [PlayOnLinux/POL-POM-5](https://github.com/PlayOnLinux/POL-POM-5) please.

PR #182, opened 2017-03-05
Allows wine to execute bat and msi files

qparis:
Allows wine to execute bat and msi files

### DIFF
--- a/Functions/Engines/Wine/script.js
+++ b/Functions/Engines/Wine/script.js
@@ -180,11 +180,11 @@ Wine.prototype.run = function (executable, args, captureOutput) {
     var extensionFile = executable.split(".").pop();
 
     if(extensionFile == "msi") {
-        return this.run("msiexec", ["/i", executable], captureOutput);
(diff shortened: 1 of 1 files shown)


Functions/Engines/Wine/script.js
```diff
@@ -177,6 +177,16 @@ Wine.prototype.run = function (executable, args, captureOutput) {
         args = [];
     }
 
+    var extensionFile = executable.split(".").pop();
+
+    if(extensionFile == "msi") {
+        return this.run("msiexec", ["/i", executable], captureOutput);
```
plata: As I said in another comment already, I'm not sure this is really a good idea. You loose completely the ability to run a msi with anything else than `/i`.

qparis: You could still use msi command.
I think I need to add the arguments

plata: True (because msiexec does not have the extension "msi"). Do you want me to merge?

qparis: Yes this can be merged if you are ok

Issue #183, opened 2017-03-06
Bat file execute not working


feanor12:
I tested the bat file launcher and it almost works.

The start command wants a wine path.
So `wine start /home/user/...LoL/run.bat` does not work, but `wine start C:\LoL\run.bat` works.

Is there a function to convert the path like `winepath -w path`?

What also works is just running the bat file(`wine start run.bat`), if I'm in the correct directory.

plata:
I don't think we have the winepath functionality already. But it should not be too difficult to implement a function in Wine/script.js which provides this (basically a wrapper for winepath).

Something like (untested):
```Javascript
Wine.prototype.winepath = function (path) {
    var winepathBinary = this._fetchLocalDirectory() + "/bin/winepath";
    var processBuilder = new java.lang.ProcessBuilder(Java.to([winepathBinary + " -w", path].concat(args), "java.lang.String[]"));
    this._process = processBuilder.start();
    return org.apache.commons.io.IOUtils.toString(this._process.getInputStream());
}
```

PR #184, opened 2017-03-07
Fix bat execution

feanor12:
* added the winepath method to Wine, which converts Unix paths to winepaths
* fixed the bat execution #183 

The bat execution works without the winepath method, but maybe it is useful for other things.
What do you think?


Functions/Engines/Wine/script.js
```diff
@@ -36,6 +36,14 @@ Wine.prototype.wizard = function (wizard) {
 };
 
 /**
+ * @param {String} [path]
+ * @returns {String}
+ */
+Wine.prototype.winepath = function (path) {
+    return this.run("winepath",["-w",path],true)
```
plata: Could you add a space after the ","?

plata:
As the winepath function is not required for the bat fix, I would prefer to split this in two pull requests (just that the changes are more visible in the commit history later on).

feanor12:
I'm bad at splitting PRs... ;) 
#187 #186

PR #194, opened 2017-04-05
Fix missing bean "repository" error

madoar:
This PR solves issue https://github.com/PlayOnLinux/POL-POM-5/issues/695.

### DIFF
--- a/Functions/Shortcuts/Wine/script.js
+++ b/Functions/Shortcuts/Wine/script.js
@@ -34,7 +34,7 @@ var WineShortcut = function () {
 
     that.miniature = function(miniature) {
       if(isArray(miniature)) {
-          var application = that._appsManager.cachedRepository().getApplication(miniature);
+          var application = that._appsManager.getApplication(miniature);
           if(application != null && application.miniatures != null && application.miniatures[0] != null) {
               that._miniature = application.miniatures[0];
           }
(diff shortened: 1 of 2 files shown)


madoar:
I've tried this on my desktop PC, but it would be nice if someone could try it on his/her PC too, before merging this.

Functions/Shortcuts/Wine/script.js
```diff
@@ -34,7 +34,7 @@ var WineShortcut = function () {
 
     that.miniature = function(miniature) {
       if(isArray(miniature)) {
-          var application = that._appsManager.getApplication(miniature);
+          var application = that._appsManager.cachedRepository().getApplication(miniature);
```
qparis: I don't like the idea of exposing the fact that you are delegating to a cached repository. This could change later and should not be part of the public API

madoar: I know what you mean. 
My problem with the current approach is, that we're mixing direct returns and callbacks. 
We use callbacks for the population of the "Apps" tab, while we're using direct returns for everything else.
To change this we should standardize the access to the applications, categories and repositories.

madoar:
Ok, I've just added a new PR https://github.com/PlayOnLinux/POL-POM-5/pull/697, where I've added two new methods `getApplication` and `getScript` to the RepositoryManager. This removes the need to expose the internal repositories from inside the RepositoryManager.

madoar:
As a note:
This PR works now only in combination with PR https://github.com/PlayOnLinux/POL-POM-5/pull/697.

PR #219, opened 2017-04-21
Add new function - setVirtualDesktop

odziom91:
setVirtualDesktop = function (width, height)
Requires two variables - width and height of window.
Script creates a new enties in Windows Registry to create new Virtual Desktop with name "Default" and set up resolution with width and height variables.

### DIFF
--- a/Functions/Engines/Wine/script.js
+++ b/Functions/Engines/Wine/script.js
@@ -633,6 +633,24 @@ Wine.prototype.setSoundDriver = function (driver) {
 };
 
 /**
+ * sets Virtual Desktop with window resolution
+ * @param width {string}, height {string}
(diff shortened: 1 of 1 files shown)


Functions/Engines/Wine/script.js
```diff
@@ -633,6 +633,24 @@ Wine.prototype.setSoundDriver = function (driver) {
 };
 
 /**
+ * sets Virtual Desktop with window resolution
+ * @param width {string}, height {string}
```
plata: Are `width` and `height` really strings?

I think it should be:
```Javascript
* @param {number} width
* @param {number} height

plata:
Do you also have a script where you needed this?

PR #222, opened 2017-04-21
Stronghold HD

odziom91:
Stronghold is a very popular RTS game developed by Firefly Studios in 2001.
This game requires to setup Virtual Desktop resolution. Without it Stronghold won't start.
This is a first version of code, so maybe someone can implement a better option. :)


Games/Stronghold/Stronghold/script.js
```diff
@@ -0,0 +1,42 @@
...
+		'1024x768',
+		'1280x720',
+		'1280x1024',
+		'1366x768',
+		'1920x1080'
+		];
+		var setupWizard = SetupWizard("Stronghold HD");
+		var selectedResolution = setupWizard.menu('Choose Virtual Desktop resolution:', resolutionModes);
+		var setupResolution = resolutionModes[selectedResolution.index];
+		switch (setupResolution) {
```
madoar: alternatively to the switch/case block you could use:
```javascript
var resolutionParts = setupResolution.split("x");
var x = Integer.parseInt(resolutionParts[0]);
var y = Integer.parseInt(resolutionParts[1]);
wine.setVirtualDesktop(x, y);
```
This approach would be shorter and better extendable if some more resolutions are to be added.

plata: Instead of 
```Javascript
var setupResolution = resolutionModes[selectedResolution.index];
```
you could also use
```
var setupResolution = selectedResolution.text;
```

odziom91: Hello!
I uploaded new version of script. Please check a code after changes.

Games/Stronghold/Stronghold/script.js
```diff
@@ -0,0 +1,26 @@
...
+		var resolutionModes = [
+	    	'800x600',
+		'1024x768',
+		'1280x720',
+		'1280x1024',
+		'1366x768',
+		'1920x1080'
+		];
+		var setupWizard = SetupWizard("Stronghold HD");
+		var selectedResolution = setupWizard.menu('Choose Virtual Desktop resolution:', resolutionModes);
```
plata: You should use the `wizard` which is passed to this function as parameter.

plata:
@odziom91 any chance you could implement my review comment? I would really like to merge this but I cannot do the change myself (do not own the game so I cannot test).

plata:
Is there anyone who can finish this? Otherwise I will close the pull request because it seems to be abandoned.

Zemogiter:
@odziom91 try replacing the whole postInstall function with this (from my Subnautica script):
	.preInstall(function(wine, wizard) {
		var screenSize = java.awt.Toolkit.getDefaultToolkit().getScreenSize();
		wine.setVirtualDesktop(screenSize.width, screenSize.height);
})

It's better than making a list of every existing resolution since it will do it for you.

plata:
Closed due to inactivity.

PR #229, opened 2017-04-22
make sure that wine prefix name is valid

### LINKED ISSUE #226: some characters seem to cause issues in prefix names
When I wrote the script for "Q.U.B.E", I had the impression that Steam doesn't like "." in the folder name. It crashed all the time with `double free or corruption`. I also noticed this error every now and then for other games. Maybe ® etc. are also bad.

We should write a function which converts any given name into a proper prefix folder name.

plata:
Remove all characters which are not:
* 0-9
* a-z
* minus
* space

Fixes #226


madoar:
I'm not sure if it's such a good idea to simply remove them. I think it would be better to replace them with some other character. I guess it's possible to find an application, whose name is made up only of "illegal" characters.

madoar:
In addition what about `/` and `\`?

plata:
Could happen if the application name is Japanese/Russian... No idea how to handle that. `/` and `\` should be removed.

madoar:
Could we use the `URLEncoder`?

plata:
Wouldn't that produce a lot of `%20` etc.?

madoar:
Is that bad?

plata:
I don't know. Not quite sure which characters can cause such problems.

plata:
Also the prefix names would not really be human readable anymore.

madoar:
I'm not sure this is a real problem if the prefix names aren't directly readable anymore. Often you get a hint, because some fragments should still stay the same.

plata:
At least the prefix name is shown in the GUI. E.g. when you uninstall. I don't think we should show "Assassin_s%20Creed_Unity" there.

madoar:
For that you can use `URLDecoder` :)

madoar:
But I agree, I'm not sure if we really want to do that, because:

1. It's not really nice, and easily forgettable, to add `URLDecoder` everywhere
2. It may confuse the user if he's looking for the prefix folder and it's named differently to what he's shown in POL 5.

plata:
Yep. Would like to hear @qparis opinion about this.

plata:
@madoar @qparis unless you have any better idea I would like to merge this.

Functions/Engines/Wine/script.js
```diff
@@ -111,7 +111,7 @@ Wine.prototype.prefix = function (prefix) {
     }
 
     // set
-    this._prefix = prefix;
+    this._prefix = prefix.replace(/[^a-z0-9_\-\ ]/gi, '');
```
madoar: Just asking: aren't there `"` missing around the regex expression?

plata: No. https://www.w3schools.com/jsref/jsref_replace.asp

madoar: Oh ok, I thought the Java `replace` method was used here, therefore I was a little confused, but it seems to be a specific javascript method instead, interesting.

madoar:
I think we can merge this

PR #231, opened 2017-04-26
Epic Games Launcher

plata:
"Testing" because of [Wine bug 41258](https://bugs.winehq.org/show_bug.cgi?id=41258)


madoar:
Is the application installed with this script usable, with the linked bug? I think this can be merged.

plata:
No. Not usable currently. Just wanted to provide the script so it doesn't get lost. Not sure which level of "broken" is acceptable for "testing".

madoar:
I think we'll need to wait for @qparis for this one, because I'm also not sure how to handle this.

plata:
Ok.

plata:
@madoar this is working now.

I do a quiet installation wiht `/q` to avoid that the launcher is started automatically. Maybe `SKIP_AUTOLAUNCH="1"` could be used to achieve this as well. As you cannot change anything during the installation, it shouldn't matter.

madoar:
This works for both the online and offline installer?

plata:
With the PR I just opened, yes. There was an issue that the `OnlineInstallerScript` could handle only `.exe` and the installer is an `.msi`.

madoar:
@plata Is this now working correctly in all cases?

plata:
Yes. Just tried both installers again.

PR #232, opened 2017-04-30
Blizzard app

plata:
The setup used a lot of resources on my PC. As soon as the installation wizard showed up, CPU was ~40%. With "staging" my PC even froze completely (could only move the mouse).


Ads20000:
@plata This sounds very similar to what I got installing Age of Empires II HD before patches in the last month or so, I wonder if there is a deeper problem (I wouldn't know though because I don't hack on Phoenicis).

plata:
As the behavior is different depending on the Wine version, I tend to blame Wine currently. However, I cannot really explain it.

madoar:
I would recommend to add a short note to the description that describes, that the installation procedure may take a while. Otherwise this can be merged.

plata:
Updated with the recent description from [WineHQ](https://appdb.winehq.org/objectManager.php?sClass=version&iId=28855) and set it to "testing". I think that's the best solution so people can try but know that it might not work properly.

madoar:
I think this can be merged.

PR #234, opened 2017-05-06
Age of Empires II HD: skip broken launcher

### LINKED ISSUE #233: Age of Empires II HD doesn't load without launcher rename
If I were a user, I'd be very confused when Age of Empires II does not launch when I ask it to. The launcher (~/.Phoenicis/wineprefix/Age of Empires II HD/drive_c/Program Files/Steam/steamapps/common/Age2HD/AoK HD.exe) has to be renamed 'Launcher.exe' (and the other 'Launcher.exe' has to be renamed 'Launcher.exe.bak' or deleted) for Age of Empires II HD to launch.

If there's not a way to script this, since it has to happen after AoE II HD has been installed, there should be instructions for doing this in the wizard?

Also, I'm worried this game plays less smoothly in Phoenicis (POL 5) as opposed to POL 4 but I'm not sure how to test this because I'm not sure AoE II HD has the option to display an FPS counter. It's a bit worrying that Java's running in the background taking up 15% of my RAM on it's own, but I guess taking up RAM isn't really that much of an issue performance-wise. Hmm.

plata:
fixes #233

### DIFF
--- a/Games/Age of Empires II HD/Steam/script.js
+++ b/Games/Age of Empires II HD/Steam/script.js
@@ -1,8 +1,18 @@
 include(["Functions", "QuickScript", "SteamScript"]);
+include(["Functions", "Filesystem", "Files"]);
 
 new SteamScript()
     .name("Age of Empires II HD")
     .editor("Skybox Labs, Hidden Path Entertainment, Ensemble Studios")
     .author("Plata")
     .appId(221380)
+    .postInstall(function(wine, wizard) {
+        // skip broken launcher by replacing it with "AoK HD.exe"
+        var installPath = wine.prefixDirectory + "drive_c/" + wine.programFiles() + "/Steam/steamapps/common/Age2HD/";
+        var launcher = installPath + "Launcher.exe";
+        cp(launcher, launcher + ".bak");
+        remove(launcher);
+        var aoe = installPath + "AoK HD.exe";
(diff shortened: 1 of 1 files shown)


madoar:
I would suggest to add the fact to you're skipping the launcher when running AoE 2 everytime your start the application to the description. Otherwise this seems ready to be merged for me.

plata:
I don't think that's possible: Skipping the launcher is depending on the script (we might have several), whereas the description belongs to the application.

madoar:
Could we maybe add a hint field to the scripts? If this hint field is defined (not null) we show a yellow message with an attention mark in POL 5 for this script, that describes what the user should be aware of when installing the script?

plata:
We could do that, but then again you could just open a `message` in the wizard directly.

madoar:
The idea with the hint field is mainly to make the user aware of possible hazards when selecting the script, especially if there are more then one possible script for an application.

plata:
I see. Not really sure if this is a good idea however (thinking about the presentation in the GUI). We can think about it but then we should open a new issue for this (it's out of scope for this PR).

madoar:
That's true, where should we open such an issue?
We can merge this PR then.

plata:
I would suggest PlayOnLinux/POL-POM-5 (after we have thought about the presentation, we can tackle the scripts part).

PR #235, opened 2017-05-13
set shortcut category

plata:


madoar:
This looks good to me. I think you can merge this, especially if this is needed to run the current version of POL 5 correctly!

PR #238, opened 2017-05-21
disable winemenubuilder

plata:
fixes PlayOnLinux/POL-POM-5#687
fixes PlayOnLinux/POL-POM-5#511


madoar:
Is our management of shortcuts complete?
Do we manage the creation, maintenance and deletion of shortcuts?

plata:
I think so, yes. But you can check again in <https://github.com/PlayOnLinux/POL-POM-5/blob/master/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java> (just look for `.desktop`).

madoar:
Looks complete to me, I think we can merge this.

PR #237, opened 2017-05-20
Warcraft III

FalseCAM:
Added Warcraft III and Warcraft III The Frozen Throne

PR #239, opened 2017-05-24
Warcraft 3 TFT preinstall check

FalseCAM:
Check and warn if Warcraft III base game is not installed while installing Warcraft III TFT.


plata:
Could you also provide a PR to fix the json as described in #237?

PR #246, opened 2017-06-02
OnlineInstallerScript: use correct file extension

plata:
required by #231


madoar:
Is it always the case, that the file contains a single `.`?
If not it would be useful to maybe split on the first `.`

plata:
Why should I split on the first `.`? I need the file extension. Maybe the URL is something `abc.def.exe`.

madoar:
You're right maybe the first `.` is not the best choice, but what do you do if you have a file with an `.tar.gz` ending for example?

plata:
I don't know. Doesn't make a lot of sense to call a `.tar.gz` with Wine anyways. You probably have to extract it first and do some custom handling, so you would have to implement something in `preInstall`.

madoar:
@plata This contains a conflict

PR #251, opened 2017-06-03
- Remove "/q" from the League of Legends installation args

### LINKED ISSUE #180: Select installation language of the League of Legends BETA Client
During the installation of the League of Legends BETA client I couldn't select the language of client. It automatically installed the english version.
I believe it would be better to give the user the choice, which language he wants to install.
Maybe it's possible to ask the user if he wants to do a fast installation or an "advanced" installation, where he can set such options?

madoar:
Fixes #180.

### DIFF
--- a/Games/League of Legends/BETA Client/script.js
+++ b/Games/League of Legends/BETA Client/script.js
@@ -20,7 +20,7 @@ new CustomInstallerScript()
     .applicationHomepage("http://leagueoflegends.com/")
     .author("Plata, feanor12")
     .installationCommand(function(wizard) {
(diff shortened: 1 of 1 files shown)


plata:
Thanks. Btw: Any particular reason why you sometimes prepend "-" to your commit messages (just interested)?

madoar:
I always write my commits as a listing.
Therefore I always prepend a "-" before each change I did in a commit ;)
If a commit only has one change it starts with a "-" too, this is easier than to first think about how many changes I introduced in a single commit.

Issue #49, opened 2017-01-03
Translate strings of the wizard


plata:
We could use <http://www.localeplanet.com/> and <https://docs.transifex.com/formats/json>.

plata:
Maybe: <https://github.com/adrienrn/jsgettext>

plata:
Another possibility would be to call `tr` directly:
```Javascript
Packages.org.phoenicis.configuration.localisation.Localisation.tr("translation test.")
```

This is working but I'm not sure how we can get the text into our `.pot` and `.po` files (especially for user repositories).

qparis:
Called tr directly makes more sense. The analysis of the scripts is another problem we can handle quite easily

plata:
If we do that, all strings have to be in the properties file which is generated with `mvn gettext:dist`. I don't see how this should happen for user repositories.

qparis:
The pot files should not necessarily be generated by maven. 
We could imagine another process that could generate all the translations from all our git repositories and uploading them to a translation platform. 
This is the way it is done in POLv4

plata:
Sure. But that's still not enough. For example, I could provide my own scripts in a separate repository. In this case, I would not have any chance to translate those.

qparis:
You could add your translation to the main repository, it should work fine

plata:
Doesn't that defeat the whole point of custom repositories? I might not even want my strings to be available publicly (e.g. in a company environment). With a solution which loads a json with the translations at runtime, I could provide my own translation file for my repository.

qparis:
Correct

qparis:
Then we should also allow to include po files in a repository

plata:
Yes.

plata:
The problem with the solution from localeplanet is that it seems to do only a simple mapping. So no plural forms etc.

plata:
I managed to get i18next running after changing the [i18next.js](https://unpkg.com/i18next/i18next.js) a little bit. So I guess we could use that.
However, there's still another problem: How do we translate the `.json` files?

plata:
I opened a PR (PlayOnLinux/POL-POM-5#870) to solve the `.json` translation).

plata:
Fixed.

PR #254, opened 2017-06-04
i18n proof-of-concept

plata:
requires PlayOnLinux/POL-POM-5#872


plata:
As the translations are stored in json, you cannot use the real string as key (e.g. no dots allowed). Also, I would like to replace `i18next.t` by `tr`. Anyone knows how to do this?

madoar:
Does `val tr = i18next.t;` work?

qparis:
I think translation context is typically a thing that should be injected from Java

qparis:
Thinking about it, I don't this we should use another i18 framework written in JavaScript.
Why not using java's tr directly?

plata:
@madoar I've tried that already. Doesn't work for some reason but I don't understand why.

plata:
@qparis I've tried to explain it in #49, but let me try again: gettext-commons creates a `Messages_lang.properties` or `Messages_lang.class` (e.g. `Messages_de.properties`) which contains the translated Strings. This is then somehow included in the jar. If we want to use `tr` in JS as well, all Strings from JS have to be in those `.properties`/`.class` files. While we could control that for our own repository, it would become practically impossible for any one else to translate their own repositories. With my proposed solution we read the translation files from the repository at runtime which solves this issue.

If you know a way to inject the translation files from the repositories into gettext-commons, that could also be a solution (I don't know if/how that could be done).

madoar:
This PR seems to have a conflict.

plata:
Outdated.

PR #256, opened 2017-06-06
German application.json for all apps

plata:


madoar:
Is it only possible to translate the description of an application or can the name also be translated with this?
I'm asking just in case we have or will have an application that has a different name in another language

plata:
You can also change the name.

PR #259, opened 2017-06-11
translate apps and categories

### LINKED ISSUE #255: translate categories
similar to application.json

plata:
requires PlayOnLinux/POL-POM-5#884

fixes #255 
fixes PlayOnLinux/POL-POM-5#879

Steps:
* Generate the `.pot`:
    ```bash
   i18n/update_translations.py
    ```
* Create a `.po`:
    ```bash
    msginit -i i18n/keys.pot -o i18n/de.po
    ```
    or update existing `.po`:
    ```bash
    msgmerge -U i18n/de.po i18n/keys.pot
    ```
* Translate
* Generate `.properties`
    ```bash
    msgcat --properties-output i18n/de.po -o i18n/Messages_de.properties
    ```


i18n/update_translations.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
```
madoar: Why are we using a python script for this?
Is it possible to use a javascript script here?

plata: Why Javascript? Can you even run it stand alone?

madoar: Yes with `nodejs`. The benefit of javascript is, that we don't need to introduce a third programming language (namely python).

qparis: With nashorn *

madoar: Depending if you want to write a java application around it, yes with nashorn ;)
If not simply install `nodejs` and run `node script.js` (if I remember correctly)

plata: I don't want to put this in a Java application. It's just a simple helper script and it was easy to implement in python (this is nothing which would usually be done in JS). I don't see a problem with a different programming language because you just have to use it (i.e. there's no need to learn python to translate your scripts).

madoar: The problem with this is, that we're basically adding another developer dependency with this.
We'll have the same problem if we require that the user needs to install `nodejs`...
I think the best way would be to use a `javascript` script, which is run with `nashorn`, which should be provided by Java 8.
After a short look on google I found [this](http://www.n-k.de/riding-the-nashorn/) where they write about `jjs test.js`.

plata: I still believe that this script isn't something anyone would write in JS. However, I also don't know enough about JS (command line args, parsing json, calling system commands etc.) to easily translate the script to JS.

qparis: Would the script run with Jython?

plata: Yes (you need Jython 2.7). But why would you want to do this?

qparis: We could have a maven goal to generate translations

madoar: The scripts are currently not "managed" by maven.
We would need to integrate it first.

plata: I'm not sure that makes sense. We do not build the scripts repository.

madoar: I think we should merge this for now, and think about a way, to integrate the functionality of the python script better in our build chain, later on.

plata: Agree. However, I don't think we should even do that. In fact this script does not really belong to this repository. It's something anyone who creates a repository will/might need.

plata:
@qparis can I also merge this?

Issue #179, opened 2017-03-04
Bugsplats during the Installation of the League of Legends BETA Script


madoar:
During the installation of the League of Legends BETA script I saw two to three bugsplats.
After the installation the client tends to freeze during the client start up and it displayed me a bugsplat the one time when I was entering the training mode.
The bugsplat when entering the training mode doesn't make the client unusable, because the BETA client seems to be able to detect a crash and automatically restarts itself afterwards which worked, but it's still quite annoying.
To resolve the other bugsplats and client freezes I killed all League of Legends processes in the Play on Linux 5 GUI and then restarted the client until it worked.

plata:
This seems to be an issue of wine rather than PoL. Can you reproduce this in plain wine?

feanor12:
I think some DLLs are missing, but I can't figure out which. I have one installation running on pol4 which works really well, but was a bit tinkered with. ;) The install on pol5 generated bugsplats, but it works sometimes. I think it has something to do with the internet connection or ssl. maybe wininet or something like that.

feanor12:
With the lol patch today it seems a bit more stable, but I also added the overrides from #189 and installed wininet.

plata:
If you can verify that any of those improves the script, feel free to enhance it ;)

feanor12:
I'm not really sure yet, as I'm stuck at the beta client loading screen. Black screen with loading animation(LoL icon + circle).

plata:
Any error output?

feanor12:
No not really. What WInedebug flags are useful? all produces a lot of output;)

plata:
I've no idea. I guess winehq is a better place for this (doesn't sound like it's something specific to PoL).

madoar:
I've found it useful and helpful to change a few things in my league of legends wine settings, that are also described in the long POL 4 League of Legends Thread history, where aztorius half a year ago wrote:

> Runs well with patch 6.21 on Fedora and Wine 1.9.19-staging.
> 
> Changes :
> 
> - upgrade Wine to 1.9.19-staging
> 
> - switch to winxp sp3
> 
> - grab mouse in fullscreen mode by default
> 
> You may need to enable CSMT in order to open League of Legends for the first time (Wine config).
> 
> Without CSMT, LoLPatcher.exe crash when downloading the game for the first time.

plata:
[...]

PR #260, opened 2017-06-16
EA Origin

plata:
* Local seems to be stuck in an endless loop during the update
* Online installs but crashes when starting it (OriginWebHelperService.exe and OriginClientService.exe)


madoar:
So both scripts are currently broken?
Should we add them or wait until they somehow work?

plata:
We should leave the PR open for now. I just wanted to share it as a base for discussion.

plata:
The legacy script is working if I do the `postInstall` steps manually. @qparis am I doing something wrong with the extract or so?

qparis:
it does not seem so

madoar:
Don't we have a similar problem with the League of Legends script as described in #179, where you can install LoL manually but not through POL5?

plata:
I just looked at the folder during the installation. It's actually working! But: The extracted files only appear in the folder after the installation is long finished. How can that possibly happen?

madoar:
Just to be sure, are we talking about the `zip` extract or another extract here?

plata:
Yes, the `zip` extract with `Extractor`. 

Just figured out that I cannot even show a message in `postInstall` (with the `wizard`). Like the installation is aborted already (maybe because there is an error in the setup?).

plata:
Ok. So the installation is not aborted. Just the wizard is closed already.

plata:
I've tracked it down: The wizard crashes/is closed when the `WineShortcut` is created. If I remove that step, everything is ok. Any ideas what could go wrong there?

madoar:
No sry :/
If I remember correctly, the creation of a `WineShortcut` relies on both java and javascript code, right?

plata:
Yes. You cannot really debug that code, that's really an issue.

madoar:
The question that I'm asking myself is, are all exceptions/errors that occur in javascript printed in the terminal/console?

plata:
I'm not sure. At least the printouts are. I simply have no idea why the wizard is closed by the shortcut creation.

madoar:
Is it possible, that the `Extractor` terminates before all files are fully extracted?

plata:
No. I just debugged a bit: The wizard is still running but not visible anymore because for some reason the tab got closed already.

madoar:
What happens if the progress reaches `100%`?

plata:
Don't know.

Anyways: `ViewLibrary#drawContent` is called. Therefore the tab is gone.

plata:
And this is triggered when the shortcut is created.

madoar:
So both scripts are working now?

plata:
No the online one is still broken (it's not the legacy version). Also, I would like to add translations and an online script for the legacy version. I only have the problem that the link is somehow strange (www.dm.origin.com/download/legacy) and cannot be downloaded by our `Downloader`.

madoar:
Ok, please give us a hands up if you think this is ready to be merged

plata:
Sure.

plata:
@qparis how can I download `www.dm.origin.com/download/legacy` with `Downloader`?

plata:
Forget it. I found the real link.

plata:
This can be merged now. The legacy versions (online + local) are working. The other one is set to "testing" (cannot work currently due to a wine bug).

PR #265, opened 2017-06-19
avoid many changes when updating translations

plata:
Previously, all messages from json where collected in a single file. As a result, changing only one message could result in a lot of unrelated changes in the .pot because the line numbers changed (see e.g. #260). The new approach creates one file per input json to solve this issue. As a result, the real changes are easier to see/review.


madoar:
So now the input json files are processed individually and the results are then concatenated?

plata:
Yes. The python script creates one `.js` per `.json`, then `xgettext` extracts the messages (Strings) to the `.pot`.

PR #274, opened 2017-06-25
Add game "Warlock - Master of the Arcane"

madoar:


plata:
Can you add the German translation?

i18n/keys.pot
```diff
@@ -17,308 +17,301 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
-#: Internet/Internet Explorer 6.0/Online/script.js:69
```
plata: why did this change so much?

madoar: I don't know I followed your steps from https://github.com/PlayOnLinux/Scripts/wiki/Translation
E.g.:
```
python i18n/update_translations.py
msgmerge -U i18n/de.po i18n/keys.pot
msgcat --properties-output i18n/de.po -o i18n/Messages_de.properties
```

plata: That's really strange. Like the script walks your directory in a different way than on my machine.

PR #277, opened 2017-06-25
sort messages for translation

plata:


madoar:
In which order should we merge the PRs best?

plata:
First this one to reduce the number of changes in your PR.

PR #278, opened 2017-06-25
Add DirectX 10

### LINKED ISSUE #276: DirectX 10
Can we add a verb for directx 10?
I need it for http://store.steampowered.com/app/342620/.
I'm not sure if this will help run the game, but at least some seems to got it working (https://www.reddit.com/r/linux_gaming/comments/3jyecw/what_are_some_of_the_best_games_you_can_run_in/)

madoar:
This PR fixes #276.
I used https://github.com/Winetricks/winetricks/blob/master/src/winetricks and https://github.com/PlayOnLinux/Scripts/blob/master/Functions/Verbs/d3dx9/script.js as a guideline.

### DIFF
--- a/Functions/Verbs/d3dx10/script.js
+++ b/Functions/Verbs/d3dx10/script.js
@@ -1,0 +1,70 @@
+include(["Functions", "Engines", "Wine"]);
+include(["Functions", "Net", "Resource"]);
+
+Wine.prototype.d3dx10 = function () {
+    var that = this;
+
+    var extractDirectXtoSystemDirectory = function(progressBar, filesToExtract, destination, pattern) {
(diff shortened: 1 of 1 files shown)


plata:
I guess we should wait until #279 is working before we merge this?

madoar:
I think we should merge this anyhow. It gives script developers a way to use directx10.
Even if #279 doesn't work, this script can still be used, right?

plata:
Yes, but how can we know if the Verb works?

madoar:
We can only know this, if we have a script that works and requires directx 10.
But still if we don't merge this, we force other developers to reimplement this later on, which for new script developers is a large barrier and may possibly discourage them.

plata:
Ok.

PR #279, opened 2017-06-25
Add game "It came from space, and ate our brains"

madoar:
This scripts requires #278.
This script currently doesn't work. The installation seems to work, but afterwards the game directly crashes without any error message when starting.
After searching a little I found [this](https://www.reddit.com/r/GameDeals/comments/1z3p2m/itch_it_came_from_space_and_ate_our_brains_100/), where someone said that the unity 3d stuff isn't supported on linux, not sure if this is the reason?


plata:
Have you tried starting it from command line with wine?
```
WINEPREFIX="/home/<username>/.Phoenicis/wineprefix/<game> /home/<username>/.Phoenicis/engines/wine/<version>/bin/wine <game>.exe
```

I have no idea where that output goes when you run from Phoenicis.

plata:
@madoar can you try if it works by now?

madoar:
I'll test it later

madoar:
@plata The script seems to work now.
But I think I've the same issue as I had in #468, where wine doesn't seem to like my multi screen setup.
When starting the game with two screens connected, the game open in fullscreen on my main screen, but the mouse position seems to be calculated over the full length (i.e. both screens).
So to select a button on the right bottom corner of the game screen, I need to move my mouse to the right bottom corner of my second (right) monitor.

I could temporarily solve the issue by unplugging my second screen, but this isn't a longterm solution.

Any ideas?

ImperatorS79:
Temporarily launch it in a virtual desktop (1920x1080). Maybe open a wine bug about multi screen config ?

madoar:
I tried to set the virtual desktop option manually.
This lead to the game opening on my second screen, while my mouse was captured on my first screen (which is also my main screen)

plata:
No idea regarding your issue. Seems to be a general Wine problem. Maybe open a bug report on winehq.org.

Please fix the Codacy issue.

PR #283, opened 2017-07-01
SteamScript: removed broken progress indication

plata:
The progress indication does not work. The installation of Call of Juarez Gunslinger Demo (#280) could not finish because it did not detect the download progress properly.


Functions/QuickScript/SteamScript/script.js
```diff
@@ -133,24 +108,16 @@ SteamScript.prototype.go = function() {
...
+    // wait until download started
+    while (!this.downloadStarted(wine)) {
         java.lang.Thread.sleep(100);
     }
 
-    // make sure download is really finished (download folder file size is not exact)
-    setupWizard.wait(tr("Please wait until Steam has finished the download ..."));
+    // make sure download is finished
     while (!this.downloadFinished(wine)) {
         java.lang.Thread.sleep(1000);
```
madoar: Why are we using here a different amount than for `while (!this.downloadStarted(wine))`?

plata: I want to be sure that we do not get stuck here. If the download is really small, it might finish very soon. Therefore, this `sleep` is short. If we wait a bit too much to figure out if the download is finished, nothing bad happens. So we can sleep longer there to reduce the polling.

PR #287, opened 2017-07-01
Tomb Raider: Anniversary

plata:


i18n/keys.pot
```diff
@@ -959,6 +954,23 @@ msgid ""
 "the player must pick up each egg in a level using a set number of tools."
 msgstr ""
 
+#: i18n/tmp/Games/Tomb Raider: Anniversary/application.js:1
```
madoar: Why `i18n/tmp/...`?

plata: That's where the strings from the json is extracted. It's nothing new in this PR.

PR #296, opened 2017-07-16
make engines first class citizens

plata:
Store everything belonging to a certain engine in the same directory tree. This makes understanding what belongs to the engine (e.g. if you want to implement a new one) much easier.
In a second step, this should be combined with a 4-level namespace such that we get:
* Functions
    * Filesystem
    * ...
* Engines
    * Wine
    * ...
* Applications
    * Games
    * ...

requires PlayOnLinux/POL-POM-5#954


plata:
@qparis I would like to also have your approval before merging this.

Issue #298, opened 2017-07-23
No Games are visible in Apps


chriamue:
When removing .Phoenicis/cache in home folder and after Refreshing Repositories in Settings,
no Games are visible in Apps anymore. Only Games from Oldwares repository are visible.
Notepad++ in Categorie Development for example is visible.

Tested on POL-POM-5 https://github.com/PlayOnLinux/POL-POM-5/commit/13e3898ef35e719ebf497793f91d68f7b042e9e7

madoar:
Can you please start POL 5 via terminal and tell us if you get any errors?

chriamue:
I don't think there are errors, but you can have a look into attached file.
[pol5-output.txt](https://github.com/PlayOnLinux/Scripts/files/1168074/pol5-output.txt)

madoar:
Can you please take a look inside the folder `/home/chriamue/.Phoenicis/cache/git1703867752` and tell us what it contains?

chriamue:
It contains the most recent scripts repository I think.
[git1703867752.txt](https://github.com/PlayOnLinux/Scripts/files/1168089/git1703867752.txt)

plata:
Can we merge this with PlayOnLinux/POL-POM-5#957?

chriamue:
Yes, it's the same problem.

plata:
Ok, then I will close this one.

PR #301, opened 2017-07-23
Fix Internet Category Name

chriamue:
Fixes https://github.com/PlayOnLinux/POL-POM-5/issues/957 and 'No Games are visible in Apps #298'


ImperatorS79:
Effectively

PR #303, opened 2017-07-24
Update Steam July 2017

chriamue:
Steam is crashing since last Steam update.
Some crashes in net.cpp as described on this website: https://forum.level1techs.com/t/wine-crashes/117565
I did test wine staging 2.12 and it is working.
Tested only the Game AOE 2.


plata:
This issue should be fixed in 2.0.2 (latest stable version). Therefore I think that this pull request will be obsolete after merging #304. Could you verify if that is correct?

chriamue:
Yes it will be obsolete. Tested Steam on 2.0.2 stable.

madoar:
I think we need to check all steam applications for their used wine version...
I believe we have some applications, that set a wine version manually...

plata:
@madoar that's correct.

PR #305, opened 2017-07-25
Update fr.po

ImperatorS79:
Sometimes I have used Steam translated sentences (as some of the strings come directly from steam). I just need some help for 18 wheels because I don't understand some expressions


plata:
@qparis can you have a look here as well?

PR #308, opened 2017-07-28
Star Wars EAW Gold Pack

ImperatorS79:
The local version often need this fix for multyplayer and FOC->https://community.pcgamingwiki.com/files/download/188-star-wars-empire-at-war-2-gb-ram-mp-lobby-fix/

But I don't know how to add it in the script...


plata:
You probably want a `postInstall` (see [wiki](https://github.com/PlayOnLinux/Scripts/wiki/script.js#advanced) step where you download the patch using the [Downloader](https://github.com/PlayOnLinux/Scripts/blob/master/Functions/Net/Download/script.js). Checkout the [Enderal script](https://github.com/PlayOnLinux/Scripts/blob/master/Games/Enderal/Steam/script.js) to get an idea how it's done.

plata:
Why did you add the patch as a new script? Wouldn't it be better to have it as part of the Steam script?

ImperatorS79:
Now, when i add a local repo with the script to test it, the testing checkbox seems to be unusable

ImperatorS79:
The Steam script doesn't need the patch, only DVD version needs it

plata:
Is your software up to date? The 4 level namespace requires an update there.

Ok, if it's only for the DVD version, then it's fine. Is it possible to install the DVD version from Phoenicis PlayOnLinux directly?

ImperatorS79:
1) I have just rebuilt to take the change into account

2) I don't understand what you mean. I'm using a local installer script to select the setup on the DVD

plata:
2. My idea is to provide a quick script which automatically runs the install from a DVD. I don't know if that's possible.

ImperatorS79:
The problem is that the way the program is installed from a DVD can change. You need to find where is located the setup.exe. In my case, there is 2 setup.exe in two directories, one for the game and one for the expansion pack. so that I'm using the LauncherEAW.exe, to access both setup.exe

plata:
Ok. Anyways this is out of scope for this PR.

ImperatorS79:
Should I open a bug for the local repo problem ?

plata:
Yes, if you have a problem open an issue.

ImperatorS79:
@plata you will have to move my script from Games/Star Wars... to the new location in Application

PR #315, opened 2017-08-02
Add a local installer for star wars bf2 (and patch 1.1)

ImperatorS79:


Applications/Games/STAR WARS Battlefront II/Local/script.js
```diff
@@ -0,0 +1,22 @@
...
+    .executable("LaunchBFII.exe")                  
+    .postInstall(function(wine, wizard) {
+        new Downloader()
+            .wizard(wizard)
+            .url("http://static.dolimg.com/mh_netstorage/lucasfilm/patches/pc/BFIIUpdateInt1_1.exe")
+            .checksum("60eaddfaba1bc71fe8bbbb560f8da229748cfaa8")
+            .to(wine.prefixDirectory + "drive_c/Program Files/LucasArts/Star Wars Battlefront II/BFIIUpdateInt1_1.exe")
+            .get();
+            
+        wine.runInsidePrefix("/Program Files/LucasArts/Star Wars Battlefront II/BFIIUpdateInt1_1.exe");
```
madoar: Just to be sure I understand the script correctly, what you're downloading and installing here is patch 1.1?

ImperatorS79: Yes, because DVD version is often 1.0.

madoar: Maybe you should add a short comment explaining that the downloaded and installed file is patch 1.1, I know that it should be clear from this PR, but if someone looks at the script later on, it won't be as clear then.

Issue #321, opened 2017-08-13
Add verb for mspatcha


ImperatorS79:
Winetricks use W2ksp4_EN.exe that can be found here [http://ftp.anadolu.edu.tr/Service_Packs/Windows_2000/W2ksp4_EN.exe](http://ftp.anadolu.edu.tr/Service_Packs/Windows_2000/W2ksp4_EN.exe)

This .exe extracts everything it contains in a i386 directory." mspatcha.dll" is located in "i386/mspatcha.dl_/"

This is needed to install Adobe Reader. (see [Adobe Reader "How To" on wine appdb](https://appdb.winehq.org/objectManager.php?sClass=version&iId=27093&iTestingId=94436))

I tried to make a verb myself, but I don't see well how to grab that specific dll, and move it in system32 directory...

plata:
It should be possible to extract the dll with `CabExtract`. Have a look at the `d3dx9` Verb as example.

PR #322, opened 2017-08-13
Added mspatcha verb

ImperatorS79:
Should solve #321


plata:
Did you try if it works with Adobe Reader?

ImperatorS79:
It works.

PR #328, opened 2017-08-16
updated Downloader with ifUpdateAvailable

plata:
requires PhoenicisOrg/POL-POM-5#979

If the download location is specified (`to()`), the new (`ifUpdateAvailable()`) method can be used to trigger a download only if the resource available via the given URL is newer than the local file.

It is currently used to avoid repeated downloads of the available Wine versions json.

### DIFF
--- a/Utils/Functions/Net/Download/script.js
+++ b/Utils/Functions/Net/Download/script.js
@@ -32,8 +32,8 @@ var Downloader = function () {
         that._localFile = localFile;
         return that;
     };
-    that.ifUpdateAvailable = function (ifUpdateAvailable) {
-        that._ifUpdateAvailable = ifUpdateAvailable;
+    that.onlyIfUpdateAvailable = function (onlyIfUpdateAvailable) {
+        that._onlyIfUpdateAvailable = onlyIfUpdateAvailable;
         return that;
     }
     that.get = function () {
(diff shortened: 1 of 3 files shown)


plata:
@qparis the last modified date reported for <http://phoenicis.playonlinux.com/index.php/wine?os=linux> is always 0. Can you change the server to return the correct date?

Utils/Functions/Net/Download/script.js
```diff
@@ -32,6 +32,10 @@ var Downloader = function () {
         that._localFile = localFile;
         return that;
     };
+    that.ifUpdateAvailable = function (ifUpdateAvailable) {
```
madoar: What exactly is the use of this function/field `ifUpdateAvailable`?

plata: Like I wrote in the description: It allows you to only download the file again if the file on the web server is newer than the one you have downloaded (i.e. it has been updated on the server).

madoar: I'm a little confused about the type of the variable `ifUpdateAvailable` that should be passed to the function.
When taking a look at normal java APIs, like https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html I normally think that methods beginning with an `if` take a lambda function,  but I believe your method takes a boolean right?

plata: Yes, it takes a boolean. Any proposal how it should be named?

madoar: Maybe something like `alwaysCheckForUpdates` or only `checkForUpdates`?

PR #341, opened 2017-08-18
updated SteamScript to cache path of manifest file

### LINKED ISSUE #331: Wine.run throws "error=7, Argument list too long"
```
at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
at jdk.nashorn.internal.scripts.Script$Recompilation$105$4307AAZ$\^eval\_.run([Engines, Wine, Engine, Object]:232)
at jdk.nashorn.internal.scripts.Script$Recompilation$104$6856$\^eval\_.programFiles([Engines, Wine, Engine, Object]:273)
at jdk.nashorn.internal.scripts.Script$Recompilation$108$1054A$\^eval\_.manifest([Engines, Wine, QuickScript, SteamScript]:40)
at jdk.nashorn.internal.scripts.Script$Recompilation$109$1606A$\^eval\_.downloadFinished([Engines, Wine, QuickScript, SteamScript]:60)
at jdk.nashorn.internal.scripts.Script$Recompilation$43$2030$\^eval\_.go([Engines, Wine, QuickScript, SteamScript]:122)
```

The strange thing is that the same command obviously worked before (e.g. in `downloadStarted`). I could imagine that this is the same as #324.

plata:
The path of the manifest file will not change during the installation of a Steam application. Therefore it is sufficient to calculate it once.

fixes #331

### DIFF
--- a/Engines/Wine/QuickScript/SteamScript/script.js
+++ b/Engines/Wine/QuickScript/SteamScript/script.js
@@ -36,7 +36,11 @@ SteamScript.prototype.gameOverlay = function(gameOverlay) {
 };
 
 SteamScript.prototype.manifest = function(wine) {
-    return wine.prefixDirectory + "/drive_c/" + wine.programFiles() + "/Steam/steamapps/appmanifest_" + this._appId + ".acf";
+    if (!this._manifest) {
+        // cache manifest path (will not change during the installation)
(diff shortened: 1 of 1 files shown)


madoar:
Are you sure this is such a large problem? I'm asking because this is simple string arithmetics, we're doing here, nothing more.

plata:
The problem is that it uses `ProcessBuilder` to get the `programFiles`. For some reason this seems to cause #331. So: No, I don't think that it is a large problem but the solution solves #331 and saves some computation time (which I would see as a nice side effect rather than the main purpose of this PR).

PR #350, opened 2017-08-19
set wineprefix path from js

plata:
fixes PhoenicisOrg/POL-POM-5#984


madoar:
With regards to your last post in PhoenicisOrg/POL-POM-5#984, is this PR fully working?

plata:
I tried:
- install app
- run app from library
- uninstall app from library

plata:
I my post, I wanted to point out that every script now is calculating the Wine prefix path. As I put a variable inside the Wine script, this is not so bad but I think it should be improved in the future.

Issue #269, opened 2017-06-24
shortcut miniature missing if app name differs


plata:
If the name inside the `script.js` is different from the application name, the shortcut miniature is missing (see e.g. "Mafia II (Demo)").

madoar:
This may be related to https://github.com/PlayOnLinux/POL-POM-5/issues/780

plata:
I'm not sure. I think we're just using the wrong name in some places. E.g. also the container will be "Mafia II Demo". In my opinion it would make more sense to use the name from the `application.json` for this.

PR #351, opened 2017-08-19
Udpated Star Wars Empire At War

ImperatorS79:
Fix missing logo


madoar:
I'm not sure I see the relation between the commit and your comment "Fix missing logo"?
Did you forget to update the logo (i.e. main.png) or are you updating the name?

ImperatorS79:
When the name in application.json and script.js differs, the logo is missing in the library tab

madoar:
This looks like a bug to me, that the logo is missing in the library tab, if the names differ.
@plata what do you think?

ImperatorS79:
Isn't it #269 ?

madoar:
You're right. Seems like I forgot that we had this issue before...

plata:
This should have been "updated" not "udpated" ;)

PR #353, opened 2017-08-19
Added The Witcher 3

ImperatorS79:
I do not own the game but I think there is enough people paying with it under wine(youtube, ...) to make a script for it


Applications/Games/The Witcher 3: Wild Hunt/Steam/script.js
```diff
@@ -0,0 +1,29 @@
...
+    //it would be better with dark ground fix -> https://bugs.winehq.org/attachment.cgi?id=58842&action=diff&context=patch&collapsed=&headers=1&format=raw 
+    .wineArchitecture("amd64")
+    .appId(292030)
+    .preInstall(function(wine, wizard) {
+        //Ensure Directx11 full features will work, and CSMT for performance
+        var path = wine.prefixDirectory + "drive_c/witcher3.reg";
+        
+        var content =   'REGEDIT4\n'                                        +
+                        '[HKEY_CURRENT_USER\\Software\\Wine\\Direct3D]\n'   +  
+                        '"csmt"=dword:1\n'                                  +
```
plata: I think we should have convenience functions for this. Compare `setSoundDriver`.

ImperatorS79: Something like this ?

plata: Sorry, like what?

ImperatorS79: And also CSMT

plata: Yes. We have to think about what is used often enough to have a function but for stuff like DirectDrawRenderer and UseGLSL we should have one (you can even change it in the GUI).

Maybe we should also have a function like
```Javascript
Wine.prototype.setRegistry = function (path, key, value) {
    var regeditFileContent =
        "REGEDIT4\n" +
        "\n" +
        path + "\n" +
        "\"" + key + "\"=" + value 
    this.regedit().patch(regeditFileContent);
    return this;
};
```
And a similar one `getRegistry`. But that should be done in a separate PR.

plata: On the other hand, we have `RegistryParser`... Just implement the functions like you suggested. I will open another issue for the general registry stuff.

ImperatorS79:
#354 take into account

PR #358, opened 2017-08-21
fixed Wine prefix path

plata:
fixes PhoenicisOrg/POL-POM-5#1013


madoar:
We should try to make these path information in a standardized way. So either all with a `/` or all without.

plata:
Yes. I was also thinking to append the "/" to `WINE_PREFIX_DIR` ("//" is not a problem). But it works now, so...

PR #359, opened 2017-08-21
Added DOOM (2016)

ImperatorS79:
I do not own the game but it has a platinum rating on the appdb


madoar:
The problem with setting possibly working scripts to testing is, how to differentiate them from script that are set to testing because they are not working scripts (the league of legends script belongs in this category)?

plata:
I don't know. But not using "testing" for untested scripts doesn't make sense either.

madoar:
That's right but we should add some kind of note field, where we can give a short description of the working state of the script. Ideally this field should be shown during the installation wizard.

plata:
That's not a bad idea. Can you open a new issue for this?

madoar:
Yes, see #361

PR #372, opened 2017-08-26
show miniature when installing an application

plata:
requires PhoenicisOrg/POL-POM-5#1061

### DIFF
--- a/Engines/Wine/QuickScript/SteamScript/script.js
+++ b/Engines/Wine/QuickScript/SteamScript/script.js
@@ -81,7 +81,9 @@ SteamScript.prototype.go = function() {
         this._executableArgs = ["-silent", "-applaunch", this._appId];
     }
 
-    var setupWizard = SetupWizard(this._name);
+    var appsManager = Bean("repositoryManager");
+    var application = appsManager.getApplication([this._type, this._category, this._name]);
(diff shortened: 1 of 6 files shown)


Engines/Wine/QuickScript/SteamScript/script.js
```diff
@@ -81,7 +81,9 @@ SteamScript.prototype.go = function() {
         this._executableArgs = ["-silent", "-applaunch", this._appId];
     }
 
-    var setupWizard = SetupWizard(this._name);
+    var appsManager = Bean("repositoryManager");
+    var application = appsManager.getApplication([this._type, this._category, this._name]);
+    var setupWizard = SetupWizardWithMiniature(this._name, application.getMainMiniature().get());
```
madoar: I think calling `get()` is dangerous here. We can't be sure, that a script provides a main miniature. If this is not the case this line will throw an exception.

madoar: This makes no sense.
If no main miniature exists for an application `Optional.empty()` is returned. 
If you call `get()` on `Optional.empty`, then `new NoSuchElementException("No value present")` will be thrown.

plata: Now I have it. Don't know. Must have done something wrong.

PR #376, opened 2017-08-28
updated Wine shortcut to specify miniature URI

plata:


Engines/Wine/Shortcuts/Wine/script.js
```diff
@@ -46,12 +46,16 @@ var WineShortcut = function () {
     };
 
     that.miniature = function(miniature) {
-      if(isArray(miniature)) {
-          var application = that._appsManager.getApplication(miniature);
-          if(application != null && application.getMainMiniature().isPresent()) {
-              that._miniature = application.getMainMiniature().get();
-          }
-      }
+        if(isArray(miniature)) {
```
madoar: So the given miniature is either an array of strings or an URI object?
I think we should add this to the function documentation

Issue #212, opened 2017-04-21
BattlenetScript


plata:
Implement BattlenetScript (similar to SteamScript).

It is possible to launch Battlenet games via
```battlenet://{ID}```
or 
```battle.net.exe --game={ID} --install```
The ID is wow, d3, wtcg, sc2....

 Determining when the download has finished could be tricky but that can be replaced with a step in the setup wizard ("press next if download finished").

ImperatorS79:
Is it really needed afterall ? Seems all Blizzard games can be downloaded from here [https://eu.battle.net/account/download]( https://eu.battle.net/account/download). There isn't a lot of blizzard games, contrary to Steam or Origin

plata:
If you can download them directly, what's the point of the Blizzard app?

ImperatorS79:
No, the Blizzard app is alaways used. For example, Hearthstone.exe is like a pre-configured version of Blizzars app that will directly download the game.

Blizzard doesn't made a lot of game, so I don't think that make a BattlenetScript is really useful

plata:
I didn't know that you could simply download the game when I wrote this. But yes, makes sense that we do not need it.

PR #381, opened 2017-08-28
Added Hearthstone

ImperatorS79:
Until BattleNetScript (#212)  is made, this can do the job.


plata:
@ImperatorS79 don't you want to give #212 a try? It really shouldn't be too difficult. Copy `SteamScript` and change the command which is used to install.

ImperatorS79:
The problem is I don't know how to convert battlenet://{ID} into something wine can understand

plata:
Checkout this: <https://www.reddit.com/r/Overwatch/comments/3qfmms/downloading_the_game_through_the_cmd_prometheus/>

ImperatorS79:
Need to find the right name for Hearthstone

plata:
Ok. I'll merge this now. If #212 works, you can update it later.

PR #386, opened 2017-08-30
updated Tomb Raider: Legend (start trl.exe directly)

plata:
crashes when starting from Steam


madoar:
Does this fix the described behavior, or was the behavior even worse before and now it "only" crashes when starting from steam?

plata:
It fixes it. If you start the executable directly, it works.

PR #390, opened 2017-09-01
Wine: ensure that only .exe are checked with ExeAnalyser

plata:


Engines/Wine/Engine/Object/script.js
```diff
@@ -200,8 +200,10 @@ Wine.prototype.run = function (executable, args, captureOutput) {
     }
 
     // do not run 64bit executable in 32bit prefix
-    if (this._architecture == "x86" && this._ExeAnalyser.is64Bits(new java.io.File(executable))) {
-        throw tr("Cannot run 64bit executable in a 32bit Wine prefix.")
+    if (extensionFile == "exe") {
+        if (this._architecture == "x86" && this._ExeAnalyser.is64Bits(new java.io.File(executable))) {
+            throw tr("Cannot run 64bit executable in a 32bit Wine prefix.")
```
madoar: You're missing a `;` here

PR #391, opened 2017-09-01
define Wine tools in JS

plata:


Engines/Wine/Tools/ConfigureWine/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var ConfigureWine = function() {
+};
+
+ConfigureWine.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

Engines/Wine/Tools/RebootWine/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var RebootWine = function() {
+};
+
+RebootWine.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

Engines/Wine/Tools/WineConsole/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var WineConsole = function() {
+};
+
+WineConsole.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

Engines/Wine/Tools/WineRegistryEditor/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var WineRegistryEditor = function() {
+};
+
+WineRegistryEditor.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

Engines/Wine/Tools/WineTaskManager/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var WineTaskManager = function() {
+};
+
+WineTaskManager.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

Engines/Wine/Tools/WineUninstaller/script.js
```diff
@@ -0,0 +1,11 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+
+var WineUninstaller = function() {
+};
+
+WineUninstaller.prototype.run = function (container) {
```
madoar: Please add a short explanation what this class does.

plata:
For the alternative way: Yes but I wanted to use the same approach as for Wine (we should also write `ShortcutReader` etc. like this) to make it easier for people who are not so familiar with JS to understand and implement their own engines.

madoar:
The problem with the current format for me is, that I don't like the `prototypes`. I'm accustomed to using a more compact way to define classes, sadly in nashorn javascript this can only be achieved by json.

plata:
We cannot use the compact form for Wine because it's too large.

madoar:
What do you mean with large?

plata:
The JS file defining Wine is very long (many functions). I think it's very hard to read in a compact form like you suggested.

madoar:
The number of functions and their size shouldn't change when using json? The only limitation is that you can't really use the keyword `this`, but if you can accept this limitation it's quite nice to work with.

plata:
It's just about the readability. With this json format, it's a lot harder to read the file because you have more levels of indentation.

madoar:
I don't think this is a big problem. In my eyes indentation helps with the readability of the code

plata:
Imagine you have 100 loops encapsulated like this:
```
for ...
    for ...
        for ...
```
At some point, you just won't be able to recognize where you are. So what you will do is, encapsulate functionality in functions. In my opinion it's the same problem for the json approach: You have already 3 levels until your function even starts. With the prototype approach it's much more visible where a function begins and ends and what belongs to a function.

madoar:
I'm not sure we will have so many layered for loops, and even if we have such a scenario, I don't really think this a problem, because a normal developer would either use a full-fledged IDE or a text editor with syntax highlighting, right? These tools normally show you quite well where you are inside the code.

plata:
It's an exaggerated example of course. If you try to write the Wine prototype in the json format, you'll see the similarity. I just wanted to point out why (in my opinion) the prototype approach is easier to read and understand especially if you do not have a lot of experience in JS.
Additionally: Even if you have no clue what this `prototype` is all about, you can directly understand that there is a constructor and some methods. So if you know a bit about programming, you will be able to use and extend the "class".

PR #395, opened 2017-09-02
added tool to open a terminal in a Wine prefix

plata:


Engines/Wine/Engine/Object/script.js
```diff
@@ -169,6 +169,14 @@ Wine.prototype.installed = function () {
 };
 
 /**
+* returns the path to the engine binary directory
+* @returns {String}
+*/
+Wine.prototype.binDir = function () {
```
madoar: I think `binDir` is a little bit misleading, because I would assume that the method returns something late a file or directory object, therefore I think `binPath` or `binaryPath` to be more fitting.

PR #398, opened 2017-09-03
Added Far cry

ImperatorS79:


ImperatorS79:
Because it is. I took the one from steam. It didn't bother me. 
So here is a new one.

PR #399, opened 2017-09-03
added Tomb Raider: The Dagger Of Xian

plata:
"testing" because it doesn't install (see PhoenicisOrg/POL-POM-5#1118). If you install it manually, it runs (with issues).


plata:
The download is just a .zip file. So you can simply extract it to the created Wine prefix and create a shortcut from the JavaFX GUI.

PR #408, opened 2017-09-08
updated Uplay (games) with Wine staging 2.16

plata:
Wine staging 2.16 fixes missing text (libcef issue)


plata:
I was also thinking about this. I'm not completely sure yet if it's good or bad. On the one hand, apps which use the latest version probably want to always use the latest version. On the other hand, they might stop working with a newer version.
As this does not only affect this PR, I suggest merging this and then open a new issue where we decide about a `NEWEST_VERSION` in the Wine js.

PR #415, opened 2017-09-13
use LATEST_DEVELOPMENT_VERSION and LATEST_STAGING_VERSION

plata:


Applications/Games/Assassin's Creed II/Steam/script.js
```diff
@@ -5,6 +5,6 @@ new SteamScript()
     .editor("Ubisoft Montreal")
     .author("Plata")
     .appId(33230)
-    .wineVersion("2.16")
+    .wineVersion(LATEST_STAGING_VERSION)
```
madoar: Can't we alternatively support the removal of `.wineVersion()` in case that the latest version should be used?
`staging` version could then be identified by the given `.wineDistribution()`.
The question is, do we have a way to deduce if a development version is wanted?

plata: I don't believe we should do that. If possible, the stable version should be used and it should be explicit if this is not possible.

madoar: Isn't it explicitly stated if the script author uses `.wineDistribution("staging")`.

plata: For staging maybe but for development not at all. I prefer that you can directly see that this script is using the latest available version.

madoar: For development we could add a new distribution mode `development`.

plata: I really don't see the point of this.

madoar: The benefit in my eyes is, that the script author can write one less for the script. I don't think it's intuitive if the author needs to add `.wineVersion(LATEST_STAGING_VERSION)` or `.wineVersion(LATEST_DEVELOPMENT_VERSION)`. I believe that a lot of people will either forget to add `.wineVersion(...)` at all, because they think that if you don't add it, the newest version is automatically taken, or will forget what constant they need to `.wineVersion`.

plata: If they forget it and it still works, they should not have added it in the first place. It's really risky to always take the newest version because there might be regressions so it could happen that a script suddenly stops working. Therefore, it should be a very conscious decision to use the latest staging/development version.

PR #416, opened 2017-09-13
use ShortcutInfoDTO

plata:


plata:
Yes, would be nice. I didn't try if it works or not.

madoar:
We could get it working if we use the nashorn equivalent of reflection: https://docs.oracle.com/javase/8/docs/jdk/api/nashorn/jdk/nashorn/api/scripting/ScriptObjectMirror.html.
If we use this, I think we can pass json to a java method and we receive a `ScriptObjectMirror` as a parameter in the java method.

plata:
I guess that would a bit too much effort, right?

madoar:
I don't know. Theoretically we could add an annotation, which we can add to a DTO class.
Afterwards we can use this annotation to "automatically" convert a ScriptObjectMirror object to a DTO object.
This could be useful for all `Builder`s we're using from inside javascript.
Do you know which builders we're using from inside javascript, except the `ShortcutDTO` builder?

plata:
No, would have to search as well. Not even sure any other builder is used.

PR #422, opened 2017-09-15
Added quartz verb

ImperatorS79:
quartz is needed for Elite:Dangerous launcher to start, see [https://bugs.winehq.org/show_bug.cgi?id=43162](https://bugs.winehq.org/show_bug.cgi?id=43162)


Engines/Wine/Verbs/quartz/script.js
```diff
@@ -0,0 +1,36 @@
...
+    remove(this.prefixDirectory + "/TMP/");
+    
+    this.regsvr32().install("quartz.dll");
+    
+    this.overrideDLL()
+        .set("native, builtin", ["quartz"])
+        .do()
+    
+    return this;
+
```
madoar: please remove this empty line

Engines/Wine/Verbs/quartz/script.js
```diff
@@ -0,0 +1,36 @@
+include(["Engines", "Wine", "Engine", "Object"]);
+include(["Utils", "Functions", "Net", "Resource"]);
+include(["Utils", "Functions", "Filesystem", "Files"]);
+
+Wine.prototype.quartz = function(){
```
madoar: Maybe add some documentation about this function/verb

ImperatorS79: I don't see well where you want some documentation. It's quite explicit and the other verbs doesn't have any "documentation"

madoar: You're basically right.
What I've meant is that you add a javadoc like comment to the head of the method.
This comment should contain the content of the verb and its source (ideally with an url).
In this case, the comment could be `Inspired from winetricks quartz` combined with the url to the source of the winetricks quartz implementation.

ImperatorS79: Is what I've just commit what you want ?

madoar: Almost, I think the link to the winetricks implementation of the script would be better, which I think is: https://github.com/Winetricks/winetricks/blob/63bc6dbe612d017a0cb6bf6e4cde265162d75bca/src/winetricks#L8523

ImperatorS79: Yes, sorry, that wasn't what I wanted to copy. I have used https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks

ImperatorS79: It's corrected now

plata:
@madoar I think @ImperatorS79 cannot merge to master so you should do it when you accept a PR.

madoar:
I didn't merge this yet, because I want to let you take a look at the PRs, because you're more knowledgeable with the scripts.
Sorry I should have written that.

plata:
If it works and the script looks ok, I think it's perfectly fine if you merge it.

PR #426, opened 2017-09-16
updated QuickScripts to get miniature from context

### LINKED ISSUE #269: shortcut miniature missing if app name differs

plata:
requires PhoenicisOrg/POL-POM-5#1131

fixes #269

### DIFF
--- a/Engines/Wine/QuickScript/InstallerScript/script.js
+++ b/Engines/Wine/QuickScript/InstallerScript/script.js
@@ -17,13 +17,7 @@ InstallerScript.prototype.constructor = InstallerScript;
 InstallerScript.prototype.go = function() {
     this._name = this._name || "Custom Installer";
 
-    var appsManager = Bean("repositoryManager");
-    var application = appsManager.getApplication([this._type, this._category, this._name]);
-    var miniature = java.util.Optional.empty();
-    if (application) {
-        miniature = application.getMainMiniature();
-    }
-    var setupWizard = SetupWizard(InstallationType.APPS, this._name, miniature);
+    var setupWizard = SetupWizard(InstallationType.APPS, this._name, this.miniature());
 
     // if no name given, ask user
     if (this._name == "Custom Installer") {
@@ -102,7 +96,7 @@ InstallerScript.prototype.go = function() {
         .prefix(wine.prefix())
         .search(this._executable)
         .arguments(this._executableArgs)
-        .miniature([this._type, this._category, this._name])
+        .miniature(this.miniature().get())
         .create();
 
     this._postInstall(wine, setupWizard);
--- a/Engines/Wine/QuickScript/InstallerScript/script.js
+++ b/Engines/Wine/QuickScript/InstallerScript/script.js
@@ -2,7 +2,6 @@ include(["Engines", "Wine", "QuickScript", "QuickScript"]);
 include(["Engines", "Wine", "Engine", "Object"]);
 include(["Utils", "Functions", "Filesystem", "Extract"]);
 include(["Utils", "Functions", "Filesystem", "Files"]);
-include(["Engines", "Wine", "Shortcuts", "Wine"]);
 include(["Engines", "Wine", "Verbs", "luna"]);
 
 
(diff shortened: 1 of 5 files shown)


Engines/Wine/QuickScript/InstallerScript/script.js
```diff
@@ -102,7 +96,7 @@ InstallerScript.prototype.go = function() {
         .prefix(wine.prefix())
         .search(this._executable)
         .arguments(this._executableArgs)
-        .miniature([this._type, this._category, this._name])
+        .miniature(this.miniature().get())
```
madoar: Is it ensured at this point that `this.miniature()` is not `empty`?

Engines/Wine/QuickScript/QuickScript/script.js
```diff
@@ -42,6 +49,21 @@ QuickScript.prototype.category = function(category) {
...
+ * @param miniature path to the miniature file
+ */
+QuickScript.prototype.miniature = function(miniature) {
+    // get
+    if (arguments.length == 0) {
+        return this._miniature;
+    }
+
+    // set
+    this._miniature = java.util.Optional.of(miniature);
```
madoar: Is it ensured at this point that `miniature` is not `null`?

Engines/Wine/QuickScript/SteamScript/script.js
```diff
@@ -146,7 +140,7 @@ SteamScript.prototype.go = function() {
         .prefix(wine.prefix())
         .search(this._executable)
         .arguments(this._executableArgs)
-        .miniature([this._type, this._category, this._name])
+        .miniature(this.miniature().get())
```
madoar: Same as above

Engines/Wine/QuickScript/UplayScript/script.js
```diff
@@ -83,7 +77,7 @@ UplayScript.prototype.go = function() {
         .prefix(wine.prefix())
         .search(this._executable)
         .arguments(this._executableArgs)
-        .miniature([this._type, this._category, this._name])
+        .miniature(this.miniature().get())
```
madoar: Same as above

Engines/Wine/QuickScript/ZipScript/script.js
```diff
@@ -70,7 +64,7 @@ ZipScript.prototype.go = function() {
         .prefix(wine.prefix())
         .search(this._executable)
         .arguments(this._executableArgs)
-        .miniature([this._type, this._category, this._name])
+        .miniature(this.miniature().get())
```
madoar: Same as above

Engines/Wine/QuickScript/InstallerScript/script.js
plata: Technically not. However, this can only happen if `getApplication` in `QuickScript` fails. As `getApplication` is called with IDs which have been passed from Java, this is practically impossible.

Engines/Wine/QuickScript/QuickScript/script.js
plata: No. But I think it is ok to get an exception if you call `miniature` with `null` as argument.

Engines/Wine/QuickScript/InstallerScript/script.js
plata: I've thought about it again. It could happen that the `ApplicationDTO` itself does not have a miniature. We do not have this case currently but it might happen. I've updated the code to handle this.

Engines/Wine/QuickScript/QuickScript/script.js
```diff
@@ -87,3 +111,22 @@ QuickScript.prototype.preInstall = function(preInstall) {
...
+QuickScript.prototype._createShortcut = function(prefix) {
+    var shortcut = new WineShortcut()
+                       .name(this._name)
+                       .type(this._type)
+                       .category(this._category)
+                       .prefix(prefix)
+                       .search(this._executable)
+                       .arguments(this._executableArgs);
+
+    if(this.miniature().isPresent()) {
```
madoar: Just asking: 
Is it possible to use lambdas in nashorn when calling java functions?
I.e. something like this:
```javascript
this.miniature().ifPresent(miniature => shortcut.miniature(miniature));
```

plata: I didn't try but aren't arrow functions ES6? So I'd rather guess it doesn't work.

madoar: You're right, arrow functions are likely to come with Java 9.
Alternatively we should be able to use `function(miniature) { return shortcut.miniature(miniature); }`.
For more details look here: https://blogs.oracle.com/nashorn/nashorn-and-lambda,-what-the-hey

plata: Yes, but honestly I don't think that's better readable than the current solution.

PR #428, opened 2017-09-17
Added Elite:Dangerous

ImperatorS79:
Launcher should work but the game fails at compiling shaders


Applications/Games/Elite:Dangerous/Steam/script.js
```diff
@@ -0,0 +1,23 @@
...
+    .author("ImperatorS79")
+    .wineVersion(LATEST_STAGING_VERSION)
+    .wineDistribution("staging")
+    .wineArchitecture("amd64")
+    .preInstall(function(wine, wizard) {
+        wine.dotnet452();
+        wine.corefonts();
+        wine.quartz();
+        wine.vcrun2012();
+        wine.windowsVersion("win7");
```
plata: Isn't this the default anyways?

ImperatorS79: As I wanted it to be as close as possible to winetricks source, the dotnet452 verb set the windows version to 2003.

plata: Ok. Didn't see that in the review. I don't think we should change the windows version from within a verb (because you cannot tell from the outside).

plata: @ImperatorS79 can you provide a PR which removes that dotnet452 changes the windows version internally (you can show a warning if it's not 2003) and then remove `wine.windowsVersion("win7");` in this PR? I would merge it then.

PR #437, opened 2017-09-24
translate script names

plata:


i18n/de.po
```diff
@@ -1412,6 +1571,14 @@ msgstr "Origin"
 msgid "Origin is EA's PC games portal."
 msgstr "Origin ist EA's Portal für PC-Spiele."
 
+#: i18n/tmp/Applications/Games/Origin/Local (Legacy)/script.js
+msgid "Local (Legacy)"
+msgstr "Lokal (Legacy)"
```
madoar: Maybe better `Lokal (Alt)`?

plata: I'm not sure. I think "Legacy" is the wording which is used officially by EA.

PR #453, opened 2017-11-06
Added secur32

ImperatorS79:
Seems to be needed for far cry 2


plata:
Have you been able test this somehow?

ImperatorS79:
The dll is copied and the override is working. (That is all winetricks verb does)

PR #454, opened 2017-11-06
Added Dragon Ball Xenoverse

ImperatorS79:
Testing because I do not own the game.


plata:
As you couldn't test the game: can you provide some evidence that it should run with that configuration (e.g. AppDB)?

ImperatorS79:
Platinum in appDB -> [https://appdb.winehq.org/objectManager.php?sClass=version&iId=31818](https://appdb.winehq.org/objectManager.php?sClass=version&iId=31818)

Wine-staging for steam

plata:
Shouldn't this work on Wine stable as well?

ImperatorS79:
I thought Steam worked only under recent wine staging version.

plata:
I think it should work. Can you try? We should prefer stable to staging/devel if possible.

ImperatorS79:
Ok, good to know.

PR #456, opened 2017-11-06
Added Far Cry 2

ImperatorS79:
Testing because I do not own the game.


plata:
As you couldn't test the game: can you provide some evidence that it should run with that configuration (e.g. AppDB)?

ImperatorS79:
Latest test was for wine 2.1 ->[https://appdb.winehq.org/objectManager.php?sClass=version&iId=14238]( https://appdb.winehq.org/objectManager.php?sClass=version&iId=14238)

ImperatorS79:
@plata Do you want me to add this as comments inside the script ?

plata:
No. I just believe that we have to be really careful about the amount of more or less untested scripts. On the one hand, it's good to provide the script so people who would like to have it can test it. On the other hand, we shouldn't have too many scripts which just don't work.

Applications/Games/Far Cry 2/Steam/script.js
```diff
@@ -0,0 +1,15 @@
+include(["Engines", "Wine", "QuickScript", "SteamScript"]);
+include(["Engines", "Wine", "Verbs", "secur32"]);
+
+new SteamScript()
+    .name("Far Cry 2")
```
plata: Shouldn't there be a ®?

Applications/Games/Far Cry 2/application.json
```diff
@@ -0,0 +1,4 @@
+{
+     "name": "Far Cry 2",
```
plata: Shouldn't there be a ®?

plata:
The other place is still missing.

PR #457, opened 2017-11-07
Added osu!

ImperatorS79:
When an online script is used, the program display an error about ssl exception... So it's a local script.


Applications/Games/osu!/Local/script.json
```diff
@@ -0,0 +1,7 @@
+{
+	"scriptName": "Local",
+	"compatibleOperatingSystems": ["LINUX"],
+	"testingOperatingSystems": [],
+	"free": false,
```
plata: shouldn't this be `true`?

PR #466, opened 2017-12-13
added Oldwares

plata:
Move all applications from [Oldwares](https://github.com/PhoenicisOrg/Oldwares) to the Scripts repository. There's really no good reason to separate them. Additionally, the maintenance is a lot easier if all scripts are located in one repository.


qparis:
I think there are still some reasons to separate them: 
* Potentially, we could have hundred and hundreds of oldware 
* AFAIK, the legal aspect of the use of oldware depends on your contry. 
* Having two default repositories ensure that we always test this behavior

plata:
- I don't see how a game from Oldwares is not a *real* game. In my opinion, Prince of Persia is at least as much a *real* game as Druid Soccer is.
- How can there be a legal aspect when we provide scripts? If there is a problem, we have the same issue with any LocalInstallerScript.
- In fact, that didn't even happen. Before I merged the 4 lvl JS namespace some minutes ago, none of the scripts in Oldwares did show up in the applications.

qparis:
All right

Issue #462, opened 2017-11-29
Error during Steam installation + missing sound


madoar:
I'm currently writing a new script for the game [Alien Swarm](http://store.steampowered.com/app/563560/Alien_Swarm_Reactive_Drop/), which can be found [here](https://github.com/madoar/Scripts/tree/alien-swarm).
When testing the new script I stumbled over multiple problems:

1. When installing steam I get the following error message:
![auswahl_020](https://user-images.githubusercontent.com/18488086/33402062-ae8c3aca-d55b-11e7-9b5a-9c35724328b5.png)
   Luckily this message is ignorable, and can be closed without terminating the installation.
2. Afterwards I've discovered, that I had no sound in the game. This lead me to the wine configuration and Regedit to take a look at the sound configuration of wine, which looks like this:
![auswahl_021](https://user-images.githubusercontent.com/18488086/33402125-deac64fa-d55b-11e7-9d93-df7f65cf7b57.png)
![auswahl_022](https://user-images.githubusercontent.com/18488086/33402126-e04fda12-d55b-11e7-98fd-c5d336b554cc.png)
   As far as I know, the wine configuration should at least show a selected driver, which in my screenshot is `none` (Kein).
   In addition, as far as I understand https://wiki.winehq.org/Sound, a registry entry `HKEY_CURRENT_USER\Software\Wine\Drivers` should exist, which should contain at least the sound driver settings.

Do you have any idea how to fix these issues?

plata:
Is there any bug report on winehq about this? At a first glance this looks like a Wine problem to me (wouldn't know how Phoenicis should cause this).

Regarding the webhelper: Can you solve it if you run steamwebhelper.exe with winxp?

madoar:
I don't know of a bug report regarding sound problems in combination with wine and alien swarm.
Maybe you could test if you stumble over the same problem? I mean the game is free and I believe you don't even need to completely install it, because the wine settings in my case show, that no drivers are available...
Regarding webhelper: Here it seems like a change to `winxp` helps.

plata:
Just tried. I have sound.

madoar:
Kk, then it seems to be my setup.
Do you have an idea how to elegantly fix the webhelper problem? I'm not sure if it's a good idea to completely set the windows version to `winxp` inside the script.

plata:
We could set webhelper to winxp in the SteamScript directly.

madoar:
[...]

PR #468, opened 2017-12-22
Update League of Legends to make it work wit the new installer

### LINKED ISSUE #464: League of Legends script doesn't work
- The installer hash are wrong.

Once i fix the installer hash, i get this: 
LoL

Thog:
- Rename "League of Legends BETA Client" to "League Of Legends" (It isn't in beta anymore)
- Update installers checksums and installation method (That should fix
any issues with the launcher)

Everything should be working (on the launche and in-game)
This fix #464

### DIFF
--- a/Applications/Games/League of Legends/Client/script.js
+++ b/Applications/Games/League of Legends/Client/script.js
@@ -2,30 +2,15 @@ include(["Engines", "Wine", "QuickScript", "CustomInstallerScript"]);
 include(["Engines", "Wine", "Shortcuts","Wine"]);
 include(["Engines", "Wine", "Verbs", "sandbox"]);
 include(["Engines", "Wine", "Verbs", "d3dx9"]);
-include(["Engines", "Wine", "Verbs", "vcrun2015"]);
 
-// Installs League of Legends and creates two shortcuts.
-//
-// One shortcut called League of Legends which starts the launcher
-// and one shortcut called BETA Client which starts the BETA client
-//
-// Important: 
-// The BETA-Client has to be installed using the launcher.
-// After the upgrade the BETA Client will do the game patching.
-///////////////////////////////////////////////////////////////////
+// Installs League of Legends
 
 new CustomInstallerScript()
-    .name("League of Legends BETA Client")
+    .name("League of Legends")
     .editor("Riot Games")
     .applicationHomepage("http://leagueoflegends.com/")
-    .author("Plata, feanor12")
-    .installationCommand(function(/*wizard*/) {
-        return {command: "msiexec", args: ["/i", "C://LoL_tmp/LoL.msi", "APPDIR=C:\\LoL"]};
-    })
-    .category("Games")
-    .wineDistribution("staging")
-    .wineVersion(LATEST_STAGING_VERSION)
-    .preInstall(function(wine, wizard) {
+    .author("Plata, feanor12, Thog")
+    .installationCommand(function(wizard) {
         // Select the region and download the setup file
         ////////////////////////////////////////////////
         var regions = ["EU West",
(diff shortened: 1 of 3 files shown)


Applications/Games/League of Legends/Client/script.js
```diff
@@ -99,36 +84,27 @@ new CustomInstallerScript()
...
 
-        wine.d3dx9();
-        wine.vcrun2015();
+        // Create run script to start the right exe
+        /////////////////////////////////////////
+        var client = wine.prefixDirectory + "drive_c/LoL/run.bat";
+        var batContent = "start C:\\LoL\\LeagueClient.exe";
+        writeToFile(client, batContent); 
     })
     .postInstall(function(wine, wizard) {
```
madoar: If `postInstall` contains no more content, I think you can try to safely remove it :)

Applications/Games/League of Legends/Client/script.json
```diff
@@ -1,5 +1,5 @@
 {
-	"scriptName": "BETA Client",
+	"scriptName": "Client",
 	"compatibleOperatingSystems": ["MACOSX", "LINUX"],
 	"testingOperatingSystems": ["MACOSX", "LINUX"],
```
madoar: If the script works you should set `testingOperatingSystems` to `[]`, otherwise the script will only be found in the applications list if  `testing` was selected.

plata:
Could you follow the naming scheme described in the (README)[https://github.com/PhoenicisOrg/Scripts/blob/master/README.md]?

madoar:
@Thog the installation fails for me with the following error message:
grafik
I believe this is the same error message as the one shown in the second image in #464.

As a note:
I've installed the `EU West` client.

Thog:
@madoar This is related to a missing openal dep on your system (lib32-openal)

madoar:
@Thog yes you're right, the missing dependency lead to the error message.
In addition I had to install the dependencies listed at https://bbs.archlinux.org/viewtopic.php?id=135032, to make the sound work correctly. I hat the sound problem for another script too (see #462).

@plata I suggest we add a `Troubleshooting` page to the Wiki, where we give a short summary about known issues and how to solve them.
One of these issues should be a missing sound driver inside the wine configuration and the shown error message in this issue.
A good info page can be found on the CrossOver homepage (https://www.codeweavers.com/support/wiki/diag). There you can find the specific commands to install the previous written libraries for multiple operating systems (https://www.codeweavers.com/support/wiki/diag/missinglibopenal, https://www.codeweavers.com/support/wiki/Diag/MissingLibAsound)

madoar:
@Thog now that the installer/updater has finished, I could test the game.
The game works, but I've discovered the following:
I'm using a multi-screen setup with two monitors. When I start the League of Legends Client the client opens in the center of the window. 
A similar effect appears when I enter a (bot) game. Upon entering a game the client opens the game screen on my main monitor, but screen gets compressed. 
It looks like the client wants to show the game window on both screens together and compresses it to be visible only on one screen. 
This can also be seen through the mouse behavior, when I click on the right side of my main screen the client will show the click on the center of the screen.

plata:
@madoar wouldn't it be sufficient to add the required libs to the dependencies of the [deb](https://github.com/PhoenicisOrg/POL-POM-5/blob/master/phoenicis-dist/src/deb/control/control)?

madoar:
@plata I'm not sure we should add this kind of dependency to the deb, because they depend on the kind of scripts you're using.
All dependencies I needed to install are only useful for wine (as far as I know). 
This makes them unnecessary for all other kinds of scripts that don't use wine.

What would be a nice idea are runtime profiles, which the user can select.
Upon selection these profiles check the system of the user for missing dependencies for the profiles usecase.
For example, a wine profile would check if all wine dependencies are installed, if some are missing it would ask the user to install them.
The question is, can we implement such profiles and do we want to do so?

plata:
@madoar that would require root permissions at runtime. I don't think we want that. However, we should not discuss this here. Can you open an issue?

madoar:
Ok now that we've merged #279, I don't see a reason to not merge this PR.
I guess that the problem described above is somehow related to my setup.

@Thog Can you please solve the conflicts of this PR? Afterwards I think we can merge this PR.

Thog:
I will try to take a look to it next week

madoar:
@Thog thank you :)
Sorry for the additional work I caused you.

Thog:
Everything should be fine now ^^

plata:
@madoar as you have requested changes: can you review again and merge if it's ok?

PR #474, opened 2018-02-14
configure Codacy

plata:


plata:
Not sure if the globals config works like this. If it does, I will add the other stuff.

plata:
According to https://support.codacy.com/hc/en-us/articles/207994335-Code-Patterns Codacy should use the .eslintrc.yml` but it doesn't seem to work (Codacy says: "We could not find a ESLint configuration file in your project root"). Any ideas?

madoar:
Does it work now after your commits?

madoar:
I'm not sure, but I guess it's possible, that codacy hasn't used your changes yet, because they are done inside a PR.
If you take a look at the top left you'll see that the `master` branch is selected, maybe that's the reason?

plata:
Ok. Let's try if you're right.

plata:
Seems to be still the same.

PR #476, opened 2018-02-16
Codacy: turn off duplication warnings for app scripts

plata:
As the application scripts all look pretty similar, we get many duplication warnings. The warnings for the Wine scripts etc. are still active as they might provide valid information.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,6 +1,10 @@
 engines:
   eslint:
     enabled: true
+  duplication:
+    enabled: true
+    exclude_paths:
+      - "Applications/**/*.js'
    
 exclude_paths:
   - '**.po'


plata:
That doesn't seem to work as expected. Maybe the exclude path doesn't work?

madoar:
I've never worked with Codacy before so I can't tell, but didn't you write in a previous issue that you weren't sure if codacy used your configuration file?

plata:
Yes but that seems to be solved now (all the "undef" warnings are gone).

madoar:
You sure that this doesn't work?
When looking at https://app.codacy.com/app/PhoenicisOrg/Scripts/issues/index I only see three issues.

In addition we should add the link to the README

plata:
I don't understand why the issues only show that. Look at files: https://app.codacy.com/app/PhoenicisOrg/Scripts/files?bid=6547561&filters=W3siaWQiOiJjb2x1bW5JbmRleCIsInZhbHVlcyI6WzNdfSx7ImlkIjoic29ydE9yZGVyIiwidmFsdWVzIjpbIkRlc2NlbmRpbmciXX0seyJpZCI6InNlYXJjaFRleHQiLCJ2YWx1ZXMiOltudWxsXX1d

plata:
If you check the complains in the Wine js, they are partly valid and partly nonsense. I'm thinking to disable duplication checks completely.

PR #484, opened 2018-02-23
added script to validate applications

plata:
The script checks:
* application contains at least one script
* category.json
* application.json
* script.json
* miniatures/main.png (must exist and be 400x300px)

### DIFF
--- a/validate.py
+++ b/validate.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+import argparse
+import json
+import jsonschema
+import os
+from PIL import Image
+import sys
+
+parser = argparse.ArgumentParser(description="Validates applications (must have an 'application.json')")
+parser.add_argument('--apps_dir',
+                    help="applications directory (default: 'Applications' directory in working directory)",
+                    default=os.getcwd() + '/Applications')
+args = parser.parse_args()
+applications_dir = args.apps_dir
+
+is_valid = True
+
+# get category directories
+category_dirs = []
+for category in next(os.walk(applications_dir))[1]:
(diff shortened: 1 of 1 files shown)


plata:
I would like to use this in Travis CI to check pull requests.

validate.py
```diff
@@ -0,0 +1,139 @@
...
+        print "invalid script.json {}: {}".format(script_json_file, validation_error)
+        is_valid = False
+
+# check miniature
+for application_dir in application_dirs:
+    main_miniature_file = application_dir + '/miniatures/main.png'
+    if os.path.isfile(main_miniature_file):
+        image = Image.open(main_miniature_file)
+        width, height = image.size
+        if width != 400 or height != 300:
```
madoar: Are you sure about that the dimensions are required?
As far as I remember, POL 5 will automatically scale the image to fit inside the correct region.
Depending on the type of selected list view this also retains the width/height proportions.
Therefore I don't think we need to validate the image dimensions.
In addition I believe we have some scripts that use slightly different dimensions. (I believe I added such an image myself a few months ago)

plata:
I would like to merge it first and add the `.travis.yml` later.

validate.py
plata: We should require the dimensions otherwise the scaled image might look very strange (even if the proportions are preserved). 
For example, an image of 40x30px will be very blurry. An image of 4000x3000px might not be recognizable because it becomes too small. If the image has different proportions than 4:3, it will look strange in the "Applications" view (e.g. 200x200px).

From what I saw, there are currently only 3 main.png with wrong dimensions (one of those with 399x299px). I will leave them as a test for Travis and then fix them.

PR #487, opened 2018-02-25
Travis CI: push translation updates (if any)

### LINKED ISSUE #480: update translation files automatically
As we do not have a build system for the scripts repository, the translation files are usually not updated. We must find a way to update them automatically.

It should be possible to do this with Travis. However, there is no easy out-of-the-box solution, so it requires a script which is called from the `.travis.yml`.

Travis should be configured such that it runs whenever anything has been committed to master.

The required steps in the script are:
1. updated the translation files (see [wiki](https://github.com/PhoenicisOrg/Scripts/wiki/Translation))
2. if anything changed, create pull request (can be done with the GitHub API:  https://developer.github.com/v3/pulls/#create-a-pull-request)

plata:
fixes #480

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
 
 script: python ./validate.py
 
-after_success: ./i18n/push_translations.sh
+after_success: $TRAVIS_BRANCH == 'master' && ./i18n/push_translations.sh
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
 
 script: python ./validate.py
 
-after_success: if [ "$TRAVIS_BRANCH" == "master" ]; then && ./i18n/push_translations.sh; fi
+after_success: if [ "$TRAVIS_BRANCH" == "master" ]; then ./i18n/push_translations.sh; fi
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
 
 script: python ./validate.py
 
-after_success: if [ "$TRAVIS_BRANCH" == "master" ]; then ./i18n/push_translations.sh; fi
+after_success: if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./i18n/push_translations.sh; fi
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


plata:
I couldn't really test this because the script will only run on master.

plata:
I would prefer not to add comments to the old part of `update_translations.py` because it's not related to this PR. If you tell me where I should add comments in the new code I will do so.

.travis.yml
```diff
@@ -3,3 +3,9 @@ before_install:
   - sudo apt-get install -y python-jsonschema python-pil
 
 script: python ./validate.py
+
+after_success: if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./i18n/push_translations.sh; fi
```
madoar: Even if it's currently quite clear what the both conditions check, but I think we should add a comment here

i18n/update_translations.py
```diff
@@ -57,3 +57,14 @@
...
 
 shutil.rmtree(out_dir)
+
+# merge .po's and update .properties
+languages = []
+for root, dir_names, file_names in os.walk(cwd + '/i18n'):
+    for file_name in fnmatch.filter(file_names, '*.po'):
+        languages.append(os.path.splitext(file_name)[0])
+for language in languages:
+    ps = subprocess.Popen('msgmerge -U i18n/{}.po i18n/keys.pot'.format(language), shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
madoar: I believe we need a comment here shortly explaining what the commands `msgmerge` and `msgcat` do in these cases

i18n/update_translations.py
```diff
@@ -57,3 +57,14 @@
 ps.communicate()[0]
 
 shutil.rmtree(out_dir)
+
+# merge .po's and update .properties
+languages = []
+for root, dir_names, file_names in os.walk(cwd + '/i18n'):
+    for file_name in fnmatch.filter(file_names, '*.po'):
+        languages.append(os.path.splitext(file_name)[0])
+for language in languages:
```
madoar: I think you should add a short comment explaining how you find the languages

plata:
Mh. That didn't work as expected. For some reason, I can't even see the log in Travis.

PR #492, opened 2018-03-01
updated Travis secure

plata:


plata:
I've re-generated the secure. Hopefully it can push now.

PR #495, opened 2018-03-02
added output to update_translations.py

plata:
This should give us an idea about what's going on in Travis.

### DIFF
--- a/i18n/update_translations.py
+++ b/i18n/update_translations.py
@@ -9,6 +9,8 @@ import subprocess
 cwd = os.getcwd()
 out_dir = cwd + '/i18n/tmp'
 
+print "write xgettext input files to {}".format(out_dir)
+
 # load all .json files
 json_file_names = []
 for root, dir_names, file_names in os.walk(cwd):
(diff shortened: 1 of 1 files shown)


plata:
Doesn't work (at least not easily). I would have to find the last run of the loop.

madoar:
why not add an additional `print "\n"` behind the loop?

plata:
I could do that but where's the difference compared to how it's now?

madoar:
It's easier readable. I also believe that it may be a better idea to print one language in each row.
This is especially a good idea, if we have a lot of supported languages.

plata:
The languages are already printed one per row. You can check the output in Travis.

madoar:
Yes I just saw it.
I guess one or more of the commands you open a subprocess for isn't executed correctly.
Is it possible to force an output for them?

plata:
I tried but it's not really helpful (the pipe is eating all the info). What we could to is split the command and try every portion until we find the issue. But I don't really want to do this on master.

plata:
It must be something with the xgettext. We could see already that the msgmerge and msgcat is working.

madoar:
You could try to setup travis for your fork? (I guess this should work)

plata:
Or just a pull request to play around until it works?

madoar:
Whatever works I guess

PR #496, opened 2018-03-02
fixed translation updates in Travis

plata:
`xgettext` on Travis doesn't accept `file` parameter for `--add-location`.


plata:
Let's see if we get some infos.

plata:
The `find` and the `sort` is working. What else could be wrong?

plata:
Seems like the `--add-location` works differently.

PR #500, opened 2018-03-03
Clean up translation process + files

### LINKED ISSUE #499: delete .po
Crowdin generates the `.properties` directly. We can delete the `.po` files and remove the `.po` and `.properties` update from `update_translations.py`.

plata:
As Crowdin can generate `.properties` directly from the `.pot`, it is not necessary anymore to update and maintain the `.po` and `.properties` files.

fixes #499


plata:
We should not push this until we are sure that Crowdin suites our needs.

plata:
I think we can merge this now.

madoar:
Is it possible to extract/generate the affected files from Crowdin later on?

plata:
Not sure if it's possible in Crowdin but there's e.g. [prop2po](https://github.com/translate/translate/blob/de383e74ef7333e0835f2ec3d87480b3c265f3dc/translate/convert/prop2po.py) ([doc](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/prop2po.html)) which could be used for this purpose.

madoar:
I also think we can merge this, but I would like to hear what @qparis thinks about this first

plata:
@madoar we can't really break anything. Even if the tool to convert .properties to .po doesn't work, we can still get the old .po from the git history.

PR #508, opened 2018-03-18
Add Warcraft III Expansion set

Grimler91:
Hi,

The existing Warcraft III and Warcraft III TFT scripts fails for me, Phoenicis/POL-POM-5 is unable to download the installers from `https://www.battle.net/download/getLegacy?product=WAR3&locale=en-US&os=WIN` and `https://www.battle.net/download/getLegacy?product=W3XP&locale=en-US&os=WIN`. I'm guessing some redirection happens that Phoenicis/POL-POM-5 can't handle. 

Downloading through wget or similar works, and uploading the files somewhere else and changing the .url() makes the existing scripts work. 

Installing and launching TFT downloads a couple of patches that updates the game to the same version as this script install, so it isn't necessary to have both `Warcraft III TFT` and `Warcraft III Expansion Set` in the repo but I haven't removed it here..

Feedback is greatly appreciated!


plata:
Script looks good. I'm a bit confused because of the download URL. There's nothing in the URL which would indicate that it's not just "plain" WC3 (without TFT). Any ideas?

Grimler91:
Yeah, I was thinking the same first, and tried to find a frozen throne url, to no avail. 
Starting the installer gives you "Warcraft III The Frozen Throne Expansion Set" though so it is WC3 The Frozen Throne. 

Might be that Blizzard's legacy games group only maintains WC3 Expansion Set (i.e. WC3 TFT), and only supply that installer (and just call it Warcraft III even though it's the expansion set).

plata:
That explanation makes perfectly sense for me. It would be nice if we could somehow verify whether or not it is true. If so, we should remove the existing WC3 and WC3 TFT scripts to avoid confusion.

Grimler91:
I did some more investigation.

When launching WC3 RoC (if installed from the existing POL script), it downloads an upgrade which installs WC3 TFT Expansion Set. If TFT is launched afterwards, it prompts for a TFT cd key.

The Expansion Set asks for both a RoC and a TFT cd key but the TFT cd key can be omitted, so people who only own RoC can still install and play the game if we replace the existing scripts. 

Conclusion: there is no need to the two existing Warcraft Scripts. I'll update this PR and remove them.

plata:
Ok, very good. Thanks for the investigation.

Applications/Games/Warcraft III Expansion Set/Online/script.js
```diff
@@ -1,13 +1,14 @@
...
+include(["Engines", "Wine", "Verbs", "corefonts"]);
 
 new OnlineInstallerScript()
-    .name("Warcraft III")
+    .name("Warcraft III Expansion Set")
     .editor("Blizzard")
-    .applicationHomepage("http://eu.blizzard.com/en-gb/games/war3/")
-    .author("FalseCAM")
-    .url("https://www.battle.net/download/getLegacy?product=WAR3&locale=en-US&os=WIN")
+    .applicationHomepage("https://eu.battle.net/shop/en/product/warcraft-iii-the-frozen-throne")
```
plata: "http://eu.blizzard.com/en-gb/games/war3/" seems to be the official Blizzard homepage for WC3 so I would prefer that instead of the shop page of TFT.

Applications/Games/Warcraft III Expansion Set/application.json
```diff
@@ -0,0 +1,4 @@
+{
+	"name": "Warcraft® III: Expansion Set",
+	"description": "The war rages on. The demonic threat has been banished from the battle-scarred fields of Azeroth, but peace is still a distant dream. The epic conflict that began in Warcraft III: Reign of Chaos continues with more units, more missions, and more explosive strategic combat."
```
plata: This is the description for TFT. Can you use a description which makes clear that the expansion set contains Reign of Chaos and TFT?

Grimler91: Thanks for the quick feedback. I can't find any generic description of the two games, I've added a sentence that the set contains both games though.

Applications/Games/Warcraft III Expansion Set/Online/script.js
Grimler91: Well, it seems to be the official homepage for WC3 RoC (they don't seem to have one for TFT for some reason). I've changed it to that page as you suggested.

Applications/Games/Warcraft III Expansion Set/application.json
```diff
@@ -0,0 +1,4 @@
+{
+	"name": "Warcraft® III: Expansion Set",
+	"description": "The war rages on. The demonic threat has been banished from the battle-scarred fields of Azeroth, but peace is still a distant dream. The epic conflict that began in Warcraft III: Reign of Chaos continues with more units, more missions, and more explosive strategic combat. <br><br>This set contains both Warcraft III: Reign of Chaos and Warcraft III: The Frozen Throne."
```
plata: Remove space before `<br><br>`.

Grimler91:
@plata extra space removed.

PR #509, opened 2018-03-19
updated Wine LATEST_DEVELOPMENT_VERSION to 3.4

plata:


plata:
Did you fall asleep on the approve button or what happened :smile:

madoar:
No, I got a connection problem/error window multiple times in a row.
Still it's interesting, that the approve got through, but the result from the server didn't...

PR #515, opened 2018-03-21
Added The Sims

Zemogiter:
Script for the first game in the series. I've tried every wine version between 2.19-staging and 3.0 and the game crashed right before the loading screen no matter what I did. Seems to be some sort of a regression in wine. It also requires the "-skip_intro" line because the intro uses some really old codec wine does not support.


Applications/Games/The Sims/Local/script.js
```diff
@@ -0,0 +1,12 @@
+include(["Engines", "Wine", "QuickScript", "LocalInstallerScript"]);
+
+new LocalInstallerScript()
+    .name("The Sims")
+    .editor("Maxis")
+    .applicationHomepage("https://www.ea.com/pl-pl/games/the-sims")
```
plata: Please use the English version of the homepage.

Applications/Games/The Sims/Local/script.js
```diff
@@ -0,0 +1,12 @@
...
+    .wineDistribution("staging")
+    .wineVersion("2.19-staging")
```
plata: "2.19-staging" is not a valid `wineVersion` (should be "2.19").

plata:
What shall we do with this? As you said, it's not usable currently.

Zemogiter:
@plata What? I did not said it was unusable. The script works but it requires a specific version of wine.

plata:
> I've tried every wine version between 2.19-staging and 3.0 and the game crashed right before the loading screen no matter what I did.

As the script is using 2.19-staging, I've assumed that the game crashes.

Zemogiter:
Yes I meant the 2.19-staging is the last version of wine this game works on.

Applications/Games/The Sims/Local/script.js
```diff
@@ -0,0 +1,12 @@
...
+new LocalInstallerScript()
+    .name("The Sims")
+    .editor("Maxis")
+    .applicationHomepage("http://www.thesims.com/")
+    .wineDistribution("staging")
+    .wineVersion("2.19")
+    .author("Zemogiter")
+    .category("Games")
+    .executable("Sims.exe", ["-skip_intro"])
+.go();
```
plata: Please indent `.go();`.

PR #523, opened 2018-03-25
Fix setting of Wine Windows version (service pack)

plata:


Engines/Wine/Engine/Object/script.js
```diff
@@ -836,6 +836,7 @@ Wine.prototype.overrideDLL = function () {
  * @returns {string|Wine}
  */
 Wine.prototype.windowsVersion = function (version, servicePack) {
+    var that = this;
```
madoar: Where do you use `that`?

plata: l. 854 ff

PR #527, opened 2018-03-25
Codacy: fix 'wizard' is defined but never used

plata:


madoar:
Why not remove unused parameters entirely?

plata:
I could do that but the `wizard` will be passed to the function and I wanted to make this clear instead of implicitly ignoring it.

PR #535, opened 2018-03-29
Add ID to json 

plata:
Specify the ID explicitly in the .json files instead of generating it implicitly in the code.

This has several advantages
* It is clear where the ID comes from (no hidden magic)
* The ID can be chosen more carefully (e.g. currently "Notepad" and "Notepad++" would have the same ID)


madoar:
Do we want camelcase ids, I would prefer them to be lowercase? Maybe that's just me...

plata:
@madoar can we merge this? If we want to use lower case, we should change that later on (after all `script.json` and `application.json` have been merged).

PR #539, opened 2018-03-31
 ESLint: force indentation with 4 spaces 

plata:


plata:
@madoar can we merge this? The stuff Codacy complains about is not really related to the changes in this pr.

PR #541, opened 2018-04-01
Add missing application.json

plata:


madoar:
Why do we need to add `application.json` files for these scripts?

plata:
It's just to be consistent. If we want to specify the `id` explicitly instead of using the one which is generated in Java, we must have these `application.json`. So e.g. if we decide to use lower case IDs, this is also the place which must be changed.

PR #542, opened 2018-04-01
Add missing script.json

plata:


plata:
Can you open a new issue to discuss this?

I've filled `free` and `requiresPatch` to be consistent. Also, I see no reason why a verb could not be non-free in principle.

PR #555, opened 2018-04-07
[Crowdin] Update translations

plata:
Enjoy!


madoar:
@plata it seems like we again have a conflict

plata:
Just realized: we get the conflicts because #500 is not merged yet... We should really merge this now.

plata:
@madoar if we merge this, we can find out whether or not we get conflicts again.

PR #562, opened 2018-04-07
Crowdin: read json directly

plata:
I don't know how the json support in Crowdin works exactly. I hope that this solves the issue that new translations produce conflicts all the time.


plata:
Wait until #500 is merged to see if this solves the issue already.

PR #568, opened 2018-04-08
ESLint: add space-before-function-paren rule

### LINKED ISSUE #565: add ESLint space-before-function-paren rule
Which style do we want?

```Javascript
function foo(arg) {
```
or
```Javascript
function foo (arg){
```

plata:
fixes #565

If you prefer to have a space before the parenthesis, I can easily change the rule.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -13,6 +13,9 @@ rules:
   keyword-spacing:
     - 2
     - {before: true, after: true}
+  space-before-function-paren:
+    - 2
+    - never
   no-trailing-spaces: 2
   indent:
     - 2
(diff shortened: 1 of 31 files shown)


madoar:
I'm not sure this really fixes #565.
The reason is, that you're changing the style for the writing of anonymous functions in this PR, while #565 addresses named functions.

These are two different things in my opinion.
For named functions I like the Java approach, i.e. `function get(args) {`.
For anonymous functions I like a space character between the keyword `function` and the beginning brace, i.e. `function (args) {`.

plata:
That's not correct. If you do not configure anything special, the rule applies to all function type. However, it can also be configured to behave differently for named/anonymous functions like you request. See https://eslint.org/docs/rules/space-before-function-paren

madoar:
Do you want to apply the rule to all function types or do you like it more to use a different behavior for named and anonymous functions?

plata:
To be honest, I don't really care. I just want a uniform style. Please have a look at my last commit. It should implement what you have suggested.

PR #574, opened 2018-04-11
Enforce aligned JSON files

plata:
Use [json-align](https://www.npmjs.com/package/json-align) to ensure that all JSON files are nicely formatted.


madoar:
Hmmm, this looks really nice visually, but I'm not sure if we're not asking to much of the developers, if we want them to format the json code like this.

plata:
I agree that this could be a problem. On the other hand, I believe that we should somehow enforce a common formatting otherwise the repository will soon have hundreds of different looking JSON files.

PR #575, opened 2018-04-12
Object oriented engine tools

plata:


plata:
That's true. However, I would like to use the same style everywhere and I think that it's more readable if you have many methods.

madoar:
Ok, I think we can merge this.

PR #577, opened 2018-04-12
Object oriented installers

plata:


madoar:
Hmmm, I see a lot of redundancy in the scripts, i.e. all scripts contain:

```javascript
/* exported Installer */
var Installer = Java.extend(org.phoenicis.scripts.Installer, installerImplementation);
```

Can we maybe add this code on the fly, and only define a json object containing the `run` method in the file?
This would also allow us to use a single file containing both `script.js` and `script.json`

plata:
We can do this but then the installers and e.g. the tools would be different. Do we want that?

madoar:
Can't we do the same for the tools too?

madoar:
This depends on what our goal is.

If we want to make it explicit what exactly happens behind the scene, then we should use `Java.extend` inside the scripts.

If we want to make the scripts short, on the point and easy to write, then I think it's better to simply provide a json object, which Phoenicis internally uses to create a new class via `Java.extend`.
The shorter form for the configure tool, would then only contain the following content: 

```javascript
{
    run: function (container) {
        new Wine()
            .prefix(container)
            .run("winecfg")
            .wait();
    }
}
```

Phoenicis would then internally load the json file, insert its content into a template string `new EngineTool() <filecontent>;` and then finally execute the `run` method.

plata:
Can we move this discussion to #1236 please?

PR #583, opened 2018-04-19
Update .properties automatically

plata:
Crowdin saves the `.po` files. Use Travis to update the `.properties` after that.

part of fix for PhoenicisOrg/phoenicis#1256

### DIFF
--- a/i18n/update_translations.py
+++ b/i18n/update_translations.py
@@ -66,7 +66,7 @@ shutil.rmtree(out_dir)
 print "\nupdate .properties for all available languages"
 languages = []
 for root, dir_names, file_names in os.walk(cwd + '/i18n'):
-    # the .po's are named lanuage.po (e.g. de.po)
+    # the .po's are named language.po (e.g. de.po)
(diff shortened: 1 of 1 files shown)


plata:
I will not fix the Codacy issue now (it's not a new problem).

i18n/update_translations.py
```diff
@@ -61,3 +61,17 @@
 ps.communicate()[0]
 
 shutil.rmtree(out_dir)
+
+# update .properties for all available languages
+print "\nupdate .properties for all available languages"
+languages = []
+for root, dir_names, file_names in os.walk(cwd + '/i18n'):
+    # the .po's are named lanuage.po (e.g. de.po)
```
madoar: `lanuage.po` -> `language.po`

plata:
Just realized that this will not work as expected: The translation files are updated when something in the .pot changes. However, this means that the .properties will not be updated when the .po are updated but only after the .pot changes again.
Therefore I believe that #584 is a better solution. I will try to provide a PR for this.

PR #585, opened 2018-04-20
Create translation file in properties format

### LINKED ISSUE #584: Better to create .properties directly?
If we would create a .properties file directly, Crowdin could output the translated .properties.

plata:
When a .properties file is used as input, Crowdin can directly output translated .properties files.
This approach avoids problems which arise when trying to generate the .properties from .po files (see #583).

Additionally, this fixes the strange way to detect if there are actually changes (i.e. not only the date has changed) and removes the need to use `shell=True` in `subprocess.Popen`.

part of fix for PhoenicisOrg/phoenicis#1256

fixes #584


plata:
I will not fix the Codacy issue (no idea how and it's not new).

PR #598, opened 2018-04-23
Add dotnet461

Zemogiter:


plata:
Please check Codacy and Travis.

Zemogiter:
Codacy complains about the registry key removal line, despite an identical line being in dotnet452. And Travis log shows no errors.

plata:
You're right about Codacy. Travis fails in the "Check json format" step. This means that your JSON is not aligned. Check "json-align" in the [documentation](https://phoenicisorg.github.io/scripts/tools/).

plata:
Is this needed for a particular script? I.e. could you test somehow that it is working?

Zemogiter:
This particular version of .NET Framework is required (alongside vcrun2017 and DXVK which are already in the repo) for Space Engineers (http://store.steampowered.com/app/244850/Space_Engineers/).
However I can't test it in POL5 because it does not show up after clicking on the icon (happens with both cli and javafx). Tried installing from source but it failed. I have tried the POL4 version and it's working.

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.dotnet40();
+    this.dotnet452()
+    this.windowsVersion("win7");
+
+    this.overrideDLL()
+        .set("builtin", ["fusion"])
+        .do();
+
+    this.run(setupFile, [setupFile, "/q", "/c:\"install.exe /q\""])
```
plata: Copy paste error: should be ".NET Framework 4.6.1" (also below)?

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.overrideDLL()
+        .set("native", ["mscoree"])
+        .do();
+
+    this.windowsVersion(OSVersion);
+    this.windowsVersion("win7");
+
+    if (OSVersion != "win2003") {
+        print(tr("{0} applications can have issues when windows version is not set to \"win2003\"", ".NET 4.5"));
```
plata: Is this really true for 4.6.1?

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -0,0 +1,57 @@
+include(["engines", "wine", "engine", "object"]);
+include(["utils", "functions", "net", "resource"]);
+include(["engines", "wine", "verbs", "luna"]);
+include(["utils", "functions", "filesystem", "files"]);
+include(["engines", "wine", "verbs", "dotnet40"]);
```
plata: I don't think all these includes are required. At least `dotnet40` and `dotnet452` should not be needed. Please include only used things.

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -0,0 +1,53 @@
...
+        .do();
+
+    this.run(setupFile, [setupFile, "/q", "/c:\"install.exe /q\""])
+        .wait(tr("Please wait while {0} is installed ...", ".NET Framework 4.6.1"));
+
+    this.overrideDLL()
+        .set("native", ["mscoree"])
+        .do();
+
+    this.windowsVersion(OSVersion);
```
plata: Why set `windowsVersion` twice here? I assume the call `this.windowsVersion("win7");` should be removed.

plata:
Is it good to just override the OS version which has been set by the user?

Zemogiter:
@plata not sure to be honest. I was following that appdb entry when doing this

plata:
I think it's better when you leave the code to got back to the initial OS version after the setup.

plata:
Now I'm completely confused: I thought you need win7 to install .NET 4.6.1?

Zemogiter:
That is true but why would anyone try to install it an anything less than win7?

plata:
I have no idea but I don't think we should prohibit it.

plata:
@Zemogiter if you change the implementation such that it works the same way as the other .NET versions, I will merge this.

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -0,0 +1,49 @@
...
+    remove(this.system32directory() + "/mscoree.dll");
+   
+    this.dotnet452()
+    this.windowsVersion("win7");
+
+    this.overrideDLL()
+        .set("builtin", ["fusion"])
+        .do();
+
+    this.wizard().wait(tr("Please wait while {0} is installed ...", ".NET Framework 4.5.2"));
```
plata: should be "4.6.1" (same below)

plata:
Please check the Codacy errors (not the first one).

PR #600, opened 2018-04-24
Fix Messages.properties for Crowdin

### LINKED ISSUE #591: fix Crowdin integration
Currently, Crowdin does not detect any string for translation (it thinks everything is context). I have contacted their support to figure out what's going on. In the worst case, we have to go back to gettext format and update the .properties in `i18n/update_translations.py` (using `msgfmt`).

plata:
Crowdin needs key=value format.

fixes #591 (I hope)


plata:
No idea how to fix the Codacy comment.

PR #602, opened 2018-04-25
Implement OOP engines

plata:


madoar:
I think the approach is not really intuitive.
Someone who doesn't know the code in depth doesn't know why the engine implementation has an internal `_java` variable.

In addition I think it's quite easy to think that a method hasn't been implemented yet for an engine, while it's in truth implemented in the `_java` object.

Another issue I have with the code is the name `_java`.
I mean the object behind it isn't a Java implementation. 
It is a Javascript implementation extending a Java interface.

plata:
The `_java` doesn't have to stay like this. It was just a quick approach for the proof of concept. We can of course find a better name for it.

Regarding your other concerns: I don't think this is an issue for the people who write engines (they should be experienced enough).
I can understand that it might not be clear for a *user* why a certain function cannot be called from `Wine` directly. However, this can easily be solved when we make `Wine` a facade for the implementation in `_java` like this:
```javascript
Wine.prototype.getAvailableVersions = function (distributionName) {
    return this._java.getAvailableVersions();
}
```

plata:
We could even say: only the `Java` script is the "Engine", the current `Wine` script is just Wine Utils`.

PR #610, opened 2018-05-05
Add application.json for Wine plugins

plata:


Engines/Wine/Plugins/application.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name"        : "Wine Plugins",
+    "id"          : "plugins",
```
madoar: `plugins` or `wine_plugins`?

plata: `plugins` is enough. The whole include is `"engines", "wine", "plugins", ...`.

PR #616, opened 2018-05-07
Update DXVK to version 0.50

ImperatorS79:


Engines/Wine/Verbs/DXVK/script.js
```diff
@@ -10,12 +10,14 @@ include(["utils", "functions", "filesystem", "files"]);
 Wine.prototype.DXVK = function () {
     print("NOTE: you need a driver that support Vulkan enough to run DXVK");
```
plata: typo: should be "that supports"

PR #632, opened 2018-05-17
Fix Verbs

plata:


plata:
Codacy issues are not new (and maybe not even valid).

PR #638, opened 2018-05-28
Changed LATEST_DEVELOPMENT_VERSION to 3.9

Zemogiter:


ImperatorS79:
The problem is there is no wine 3.9 yet -> [http://wine.playonlinux.com/binaries/linux-x86/](http://wine.playonlinux.com/binaries/linux-x86/)

Zemogiter:
@ImperatorS79 well that is weird. One would think that if they made a change log they also uploaded a compiled package. Can't we use the source tar.xz file?

plata:
@Zemogiter it's not on the WineHQ site. It's in the PlayOnLinux Wine build service. @qparis can you please check that (3.0.1 is missing as well).

plata:
@Zemogiter bad timing :P. Wine 3.9 is available now.

PR #639, opened 2018-05-28
Update SteamScript with new sha1sum

Zemogiter:
Valve changed their SteamInstaller.exe file so the checksum is no longer correct and every steam game instalation crashes.


plata:
Thanks.

Just two remarks:
* Please use the PR title format as described in the README (I will update it).
* It's the sha1sum not the md5sum.

PR #644, opened 2018-05-30
Fix Wine tools

plata:


Zemogiter:
Why the .wait(); line was removed? Won't it cause the script to go to the next step(s) too early?

plata:
The wait is a parameter of `run`.

PR #647, opened 2018-06-07
Add Totally Accurate Battlegrounds

plata:
runs but performance could be better


Zemogiter:
Have you tried Vulkan?

plata:
No. How would I do that?

Zemogiter:
Use dxvk verb.

plata:
Just using the verb will not make the game run in Vulkan, will it?

ImperatorS79:
Pol version of wine does not support vulkan atm, so impossible to test.
[DXVK](https://github.com/doitsujin/dxvk) is a "translation layer" from d3d11 to vulkan. The verbs install 2 ddl's (d3d11 and dxgi) inside wine windows directory and set an override. A lot of players actually use it to have better performance in comparaison to wine opengl to d3d11 (just check youtube to see ;-) ).
But Totally Accurate Battlegrounds use DirectX 9, so the corresponding project is [VK9](https://github.com/disks86/VK9).

plata:
I don't think it's a good idea to use vulkan for this then. Maybe TABS will get some optimization to run better.

PR #648, opened 2018-06-12
Update Wine LATEST_DEVELOPMENT_VERSION to 3.10

plata:


Zemogiter:
Wine 3.9 and 3.10 are already in POL5 and POL4

PR #649, opened 2018-06-12
Update Internet Explorer 6.0

plata:
fix installation

### DIFF
--- a/Applications/Internet/Internet Explorer 6.0/Online/script.js
+++ b/Applications/Internet/Internet Explorer 6.0/Online/script.js
@@ -12,7 +12,7 @@ include(["engines", "wine", "verbs", "msls31"]);
 var installerImplementation = {
     run: function () {
         var appsManager = Bean("repositoryManager");
-        var application = appsManager.getApplication(["Applications", "Internet", "Internet Explorer 6.0"]);
+        var application = appsManager.getApplication([TYPE_ID, CATEGORY_ID, APPLICATION_ID]);
         var setupWizard = SetupWizard(InstallationType.APPS, "Internet Explorer 6.0", application.getMainMiniature());
 
         setupWizard.presentation("Internet Explorer 6.0", "Microsoft", "http://www.microsoft.com", "Quentin PÂRIS");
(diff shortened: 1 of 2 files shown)


Zemogiter:
Why haven't you added those fixes to my PR or tell me to do it? I mean making another PR when another PR related to the same script is open does not make sense.

plata:
It's not the same script. Typo...

Applications/Internet/Internet Explorer 6.0/Online/script.js
```diff
@@ -12,7 +12,7 @@ include(["engines", "wine", "verbs", "msls31"]);
 var installerImplementation = {
     run: function () {
         var appsManager = Bean("repositoryManager");
-        var application = appsManager.getApplication(["Applications", "Internet", "Internet Explorer 6.0"]);
+        var application = appsManager.getApplication([TYPE_ID, CATEGORY_ID, APPLICATION_ID]);
```
madoar: These lines look like black magic for everyone who doesn't know where the ids come from and what they mean.

Zemogiter: TYPE_ID refers to either Applications,Engines or Verbs. CATEGORY_ID refers to what inside the aforementioned folders. APPLICATION_ID is self-explainatory. At least thats how I understand it.

madoar: Yes, that is correct.

The issue with these fields is just that a first time script developer, who looks at this script for reference purposes for example, is likely to stumble over these variables.
The reason being is, that they are set from outside (i.e. the Java part of Phoenicis) and that they  are uncommented (i.e. https://phoenicisorg.github.io/scripts contains no explanation about these constants).

I don't think we can improve the first issue, therefore this comment is meant as a kind of note for the future, meant for us, to think about a more intuitive/understandable solution, how to make these constants accessible.

plata: This issue will be solved with https://github.com/PhoenicisOrg/phoenicis/issues/1242.

PR #650, opened 2018-06-13
Add Vk9 verb

ImperatorS79:
I need to copy some files next to the executable. I think the best way to do it is to ask the user for the path (with a directory chooser), but I do not know how to do this.


plata:
Please check why the Travis build fails.

ImperatorS79:
I have no idea why. The travis log end with : 
` diffs=$(git diff --name-only | wc -l); if [ $diffs != 0 ]; then exit 1; fi`
And then nothing.

plata:
Looks like you didn't run `json-align` as described in the README.

ImperatorS79:
Ok, I did not click on the link, sorry.

ImperatorS79:
How to install json-align on ubuntu ?

ImperatorS79:
I do not understand why it is failling.

ImperatorS79:
@plata what is the purpose of that command in the check json stage ?
`diffs=$(git diff --name-only | wc -l); if [ $diffs != 0 ]; then exit 1; fi`

plata:
It ensures that there are no changes after running the tools like json-align, so it checks that the files are formatted correctly etc.

ImperatorS79:
Ok, so I would like to use ` this.wizard().browse()` to select the directory, but it can only return a path with the .exe (exemple : I want `/home/test/mydir` but I can only select `/home/test/mydir/myexe.exe`

plata:
That's because you explicitly tell it to do so:
```Javascript
this.wizard().browse("Please select the .exe directory", this.prefixDirectory(), ["exe"]);
```

ImperatorS79:
If I do 
```javascript
this.wizard().browse("Please select the .exe directory");
```
I still cannot choose a directory and I begin in th /home/test directory, instead of the prefix one

plata:
Just checked: in `StepRepresentationBrowse`, we have a `FileChooser`. To select a directory, we need a `DirectoryChooser`. Makes sense to have it, the only question is how the API should look like: additional parameter or new method (e.g. `browseDir`). 

see #652

plata:
I guess this is blocked until we have implemented `browseDir`, right?

ImperatorS79:
Yes indeed.

mihawk90:
This is a pretty roundabout way of doing it, but you might be able to just cut off the last part of the path, no? You could slice the path at `\` (or `/`, whatever `browse()` returns), pop the last element, and then join them back together:

```JS
var pathDir = userFilePath.split('\\'); // or /
    pathDir.pop();
    pathDir = pathDir.join('\\');
```

There's probably easier ways to do it, but that would probably work.

plata:
I think the best way to do this is simply passing a parameter to the `Vk9` function. The parameter can be set to a specific version or left away to use the latest version by default.

mihawk90:
I don't know how the whole system works yet so.. yeah :D 

But I guess that is just a sort-of-internal function anyway right, I assume it gets called by the installer scripts? The user can be asked there I imagine (or later on in the configuration).

plata:
Exactly, the parameter would be passed from the installer script so this is where the dialog must be opened. That way every script can decide whether or not it's required to ask the user.

plata:
@ImperatorS79 can you implement @mihawk90's suggestion to make the version configurable?

ImperatorS79:
I'll try.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+* Setup VK9-> https://github.com/disks86/VK9
+* @param {String} vk9Version VK9 version to install
+* @returns {Wine} Wine object
+*/
+Wine.prototype.VK9 = function (vk9Version) {
+    print("NOTE: you need a driver that supports Vulkan enough to run VK9");
+    print("NOTE: wine version should be greater or equal to 3.5");
+    print("NOTE: work from 0.28.0");
```
plata: works

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -0,0 +1,59 @@
...
+
+    if (typeof vk9Version === 'undefined')
```
plata: probably better: `typeof vk9Version === 'string'`

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -0,0 +1,59 @@
...
+        .wizard(this.wizard())
+        .archive(setupFile32)
+        .to(this.prefixDirectory() + "/TMP32/")
+        .extract();
+
+    cp(this.prefixDirectory() + "/TMP32/" + vk9Version + "-bin-x86-Release/" + "d3d9.dll", this.system32directory());
+
+    remove(this.prefixDirectory() + "/TMP32/");
+
+    if (this.architecture() == "amd64")
```
plata: `===`

Engines/Wine/Verbs/VK9/script.json
```diff
@@ -0,0 +1,12 @@
+{
+    "scriptName"                 : "VK9",
+    "id"                         : "vk9",
+    "compatibleOperatingSystems" : [
+        "LINUX"
+    ],
+    "testingOperatingSystems" : [
+        "LINUX"
```
plata: "testing" still true?

Engines/Wine/Verbs/VK9/script.js
ImperatorS79: I want that code to be executed if no variable is sent. If I do that, ```vk9version``` will never be taken into account.

plata: The problem is if someone passes a number or something else which is invalid. Therefore I think it would be better to test for string and use the default otherwise.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -0,0 +1,59 @@
...
+* Setup VK9-> https://github.com/disks86/VK9
+* @param {String} vk9Version VK9 version to install
+* @returns {Wine} Wine object
+*/
+Wine.prototype.VK9 = function (vk9Version) {
+    print("NOTE: you need a driver that supports Vulkan enough to run VK9");
+    print("NOTE: wine version should be greater or equal to 3.5");
+    print("NOTE: works from 0.28.0");
+
+    if (typeof vk9Version != ='string')
```
plata: use `!==` (the space in between looks a bit strange)

plata:
Anything you'd like to change? Otherwise I think this can be merged.

ImperatorS79:
No, nothing to change.

PR #656, opened 2018-07-08
Add Space Colony

Zemogiter:
Ready for review.

PR #658, opened 2018-07-11
Update travis.yml (printing changed json files)

Zemogiter:
After the `-git diff` line is added, Travis no longer crashes on checking json format. This error happens even if the .json files are correct.


plata:
Are you sure that this really fixes the issue in #656? The checks in #661 were successful without this "fix".

PR #665, opened 2018-08-25
Update Wine LATEST_DEVELOPMENT_VERSION to 3.15

plata:


Zemogiter:
The latest version is now 3.15 and it fixes a very serious issue that makes dotnet4x installation hang at some point. But we have to wait until #668 is fixed.

PR #667, opened 2018-08-29
Fix SteamScript does not show "close Steam" message

plata:


Zemogiter:
Have you tested this fix in any steam game?

plata:
I would like to merge this PR first because it's actually a different issue then in your PR.

PR #672, opened 2018-09-05
Fix custom installer scripts

plata:


Zemogiter:
What does this fix?

plata:
The Wine distribution/version selection was broken and the custom installer crashed.

PR #686, opened 2018-09-23
Update Subnautica to use Wine 3.16, forcing DX11

Zemogiter:
I did some more testing and found out installing d3dx11 (either via winetricks or using our verb) and writing "-force-dx11" (without quotes) in the game's Launch Options increases performance even when inside large seabases on when looking at wrecks/seabases. According to my tests those things put the most stress on GPU and CPU (trying to overclock my i7-7700K to 5 GHz or somewhere close might help a little but I don't have a good enough cooler to test that). This also fixes player still moving after releasing movement keys.
Another option would be to use -force-glcore45 or -force-opengl but I found out that both options, while having more consistent framerate introduce major z-fighting on Aurora and Cyclops submarine alongside the in-game fps debug command not displaying anything. They also remove all text from F3 and F5 debug menus. Therefore using OpenGL is not an option.
Using Vulkan crashes the game on either loading screen or in game options (tested on Wine Beta/Proton 3.7). 
In conclusion the only viable way to run this game right now is via wine's implementation of DirectX11 even if it still have some shortcomings.
Are these includes still required? Due to my slow internet connection and the script sometimes fails to detect that the steam game download/installation (mostly when the game files are already there) have been finished I can't test that.
```
include(["utils", "functions", "apps", "resources"]);
include(["utils", "functions", "filesystem", "files"]);
include(["engines", "wine", "engine", "object"]);
```


Zemogiter:
Yes I'm aware we don't have `d3dx11` verb. I'm making one right now.

Applications/Games/Subnautica/Steam/script.js
```diff
@@ -2,12 +2,11 @@ include(["engines", "wine", "quick_script", "steam_script"]);
...
 include(["engines", "wine", "verbs", "vcrun2013"]);
-include(["engines", "wine", "verbs", "vcrun2008"]);
-include(["engines", "wine", "verbs", "xact"]);
+include(["engines", "wine", "verbs", "d3dx11"]);
 include(["engines", "wine", "verbs", "corefonts"]);
-include(["utils", "functions", "apps", "resources"]);
-include(["utils", "functions", "filesystem", "files"]);
-include(["engines", "wine", "engine", "object"]);
+//include(["utils", "functions", "apps", "resources"]);
+//include(["utils", "functions", "filesystem", "files"]);
```
plata:
```
include(["utils", "functions", "apps", "resources"]);
```
is still used in `AppResource().application`. The other two can be removed.

Applications/Games/Subnautica/Steam/script.js
```diff
@@ -17,22 +14,21 @@ var installerImplementation = {
             .author("Zemogiter")
             .applicationHomepage("https://unknownworlds.com/subnautica/")
             .wineDistribution("upstream")
-            .wineVersion(3.9)
+            .wineVersion("3.16")
```
plata: Do you want 3.16 explicitly or can you use the latest development version as well?

Zemogiter: There is a chance the next development version might introduce some regressions to this game. I'd rather test it beforehand. If it works then someone will make a PR.

plata: Ok.

PR #660, opened 2018-07-11
Update DXVK to always use the latest release

Zemogiter:

PR #694, opened 2018-10-10
Update DXVK to version 0.81

ImperatorS79:
Note that DXVK now supports also d3d10.


plata:
Doesn't this conflict with #660?

ImperatorS79:
I don't think so if you merge this first

plata:
Ok.

PR #697, opened 2018-10-12
Update DXVK (install .dll's in correct directories)

ImperatorS79:


plata:
Is this really a workaround for a bug or not rather something like "Update DXVK (install .dll's in correct directories)"?

ImperatorS79:
You are absolutely right

PR #703, opened 2018-11-03
Update Steam URL and Checksum

mlaradji:
Two changes were made:
1. Update the out-of-date SHA1 checksum of SteamSetup.exe.
2. Change HTTP links to HTTPS.

### DIFF
--- a/Applications/Games/Steam/Online/script.js
+++ b/Applications/Games/Steam/Online/script.js
@@ -5,10 +5,10 @@ var installerImplementation = {
         new OnlineInstallerScript()
             .name("Steam")
             .editor("Valve")
-            .applicationHomepage("http://www.steampowered.com")
+            .applicationHomepage("https://www.steampowered.com")
             .author("Quentin PÂRIS")
-            .url("http://media.steampowered.com/client/installer/SteamSetup.exe")
-            .checksum("e930dbdb3bc638f772a8fcd92dbcd0919c924318")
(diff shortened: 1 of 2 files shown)


plata:
Can you update the QuickScript as well?
https://github.com/PhoenicisOrg/scripts/blob/bb5d7e4d0f219e48528baa5abf59f715ece3a9a4/Engines/Wine/QuickScript/Steam%20Script/script.js#L110

plata:
Did you test this? Just wondering because the checksum seems to be unchanged.

Zemogiter:
Looks like the checksum for Steam Script was updated before.

mlaradji:
The issue I faced was a "Checksum error" that occurs when downloading Steam. This is now fixed.

@plata. I've tested this, and it seems to be working fine. The checksum in `Engines/Wine/QuickScript/Steam Script/script.js` was indeed already up-to-date, but the checksum in `Applications/Games/Steam/Online/script.js` was not.

I have some reservations about using `https://steamcdn-a.akamaihd.net/client/installer/SteamSetup.exe` instead of `http://media.steampowered.com/client/installer/SteamSetup.exe`, as the former looks like it can be changed. Given that Phoenicis validates the exe using the checksum, a corrupt installation file won't be a problem, and so the only real benefit of using HTTPS is privacy.

plata:
Let's try it. If it changes too often, we can think again.

PR #709, opened 2018-11-19
Update Wine (fix "content server unavailable")

plata:
TODO: remove when fix is available in Wine stable

### DIFF
--- a/Applications/Games/Steam/Online/script.js
+++ b/Applications/Games/Steam/Online/script.js
@@ -1,4 +1,5 @@
 include(["engines", "wine", "quick_script", "online_installer_script"]);
+include(["utils", "functions", "filesystem", "files"]);
 
 var installerImplementation = {
     run: function () {
@@ -11,6 +12,7 @@ var installerImplementation = {
             .checksum("4b1b85ec2499a4ce07c89609b256923a4fc479e5")
             .category("Games")
             .executable("Steam.exe", ["-no-cef-sandbox"])
+            .wineVersion(LATEST_DEVELOPMENT_VERSION) // fixes "content server unavailable" error (Wine bug 45329), TODO: remove when fix is available in Wine stable
             .go();
     }
 };


plata:
I couldn't test this. @mlaradji @suomilanittaja can you try if it solves the issue?

suomilanittaja:
tested this and it not worked. broke more thinks like friends and chat not opening anymore

suomilanittaja:
@mlaradji not worked

plata:
@mlaradji the workaround you linked there should be what is implemented for the Steam QuickScript. I also saw that the bug is supposed to be fixed. That's why I'm using the latest Wine devel version in this pull request. However, this now seems to suffer from https://bugs.winehq.org/show_bug.cgi?id=45948.

Like I explained in the linked issue, it is not possible to implement the config workaround for this script because it simply does not exist during installation.

I don't currently see a possible fix for this on our side (all the issues need to be fixed in Wine, there's nothing we could do about that).

In any case, I'm not really using the Steam app myself. Whenever I want to install a Steam game, I quickly write a script for it (it's really super fast). That way, I can also use the workaround implemented in the QuickScript. Additionally, other users who would like to install the same game can use it later on as well.
Therefore, I would like to encourage you to also try this approach (at least until Wine is fixed again).

mlaradji:
Given the huge variety of Windows games on Steam, I agree with @plata's approach that one should implement a script for each individual game, rather than attempt to have one fully-working Steam script.

Also, depending on the game, you might be able to use the native Linux Steam app. If you opt-in for the Steam Client Beta, there is an in-client option to use Wine-based `Proton` (https://github.com/ValveSoftware/Proton/) to attempt to run Windows games on Linux. I have had mixed results with `Proton`, however, but this approach would at least allow you to download Windows games on Linux systems, after which one can use their own configuration (e.g. on Wine) to run it.

plata:
I will close this PR because the "fix" doesn't work.

PR #718, opened 2018-12-06
Update UplayScript, added message regarding winbind

Zemogiter:


Engines/Wine/QuickScript/Uplay Script/script.js
```diff
@@ -54,6 +54,7 @@ UplayScript.prototype.go = function () {
         .prefix(this._name, this._wineDistribution, this._wineArchitecture, this._wineVersion)
         .luna();
 
+    setupWizard.message(tr("Please install winbind before installing Uplay."));
```
plata: Maybe better something like "Please ensure that winbind is installed before you continue."?

Zemogiter: Sounds like the same thing just with different words.

plata: I just want to avoid misunderstandings.
E.g.:
- If I have already installed winbind, why is Phoenicis asking me to install it?
- A user might assume that winbind gets installed when he clicks "Next".

PR #722, opened 2018-12-10
Update Starcraft 2 (use default Windows version and Wine 3.19-staging)

Zemogiter:
The game was tested in 5 different co-op matches and it did not crash.


ImperatorS79:
You should remove testing in the script.json then ^^.

PR #729, opened 2018-12-12
Update Subnautica (change Wine version, use DXVK)

Zemogiter:
Another performance update to this script. Tested and ready for review.


Applications/Games/Subnautica/resources/fix.reg
```diff
@@ -3,8 +3,8 @@ REGEDIT4
...
-[HKEY_CURRENT_USER\Software\Wine\AppDefaults\Steam.exe]
-"Version"="winxp64"
- 
-[HKEY_CURRENT_USER\Software\Wine\AppDefaults\steamwebhelper.exe]
-"Version"="winxp64"
+[HKEY_CURRENT_USER\Software\Wine]
+"HideWineExports"="Y"
+
+[HKEY_CURRENT_USER\Software\Wine\Direct3D]
+"csmt"="0x00000000"
```
plata: Use the csmt plugin instead.

Applications/Games/Subnautica/resources/fix.reg
```diff
@@ -3,8 +3,8 @@ REGEDIT4
 [HKEY_CURRENT_USER\Software\Wine\X11 Driver]
 "UseTakeFocus"="N"
 
-[HKEY_CURRENT_USER\Software\Wine\AppDefaults\Steam.exe]
-"Version"="winxp64"
- 
-[HKEY_CURRENT_USER\Software\Wine\AppDefaults\steamwebhelper.exe]
-"Version"="winxp64"
+[HKEY_CURRENT_USER\Software\Wine]
+"HideWineExports"="Y"
```
plata: What does this do? Should we have a plugin for it?

Applications/Games/Subnautica/resources/fix.reg
Zemogiter: CSMT was used to increase performance in some applications. Its now depreciated.

Applications/Games/Subnautica/resources/fix.reg
Zemogiter: It hides wine version from appliactions in wineprefix. It was needed at some point before DXVK became a viable option. Now I've tried to run Subnautica without it and it made no difference. Very few games need wine-staging and of those games maybe one or two really needs this setting to be enabled. We can always make a plugin for that but it's far from being a top priority.

plata: I see. But if the script doesn't need it, we should just remove it.

PR #731, opened 2018-12-13
Implement OOP Verbs

plata:
requires https://github.com/PhoenicisOrg/phoenicis/pull/1569

I didn't check all the Verbs. I validated that the concept is working and that the Verbs which require special input (DXVK, VK9, sp3extract) use the wizard to retrieve the needed user input.

### DIFF
--- a/Engines/Wine/Verbs/DXVK/script.js
+++ b/Engines/Wine/Verbs/DXVK/script.js
@@ -4,8 +4,9 @@ include(["utils", "functions", "net", "resource"]);
 include(["utils", "functions", "filesystem", "files"]);
 
 /**
-* Setup DXVK-> https://github.com/doitsujin/dxvk/
-* @param dxvkVersion DXVK version to download
+* Verb to install DXVK
+* see: https://github.com/doitsujin/dxvk/
+* @param {String} dxvkVersion DXVK version to download
 * @returns {Wine} Wine object
 */
 Wine.prototype.DXVK = function (dxvkVersion) {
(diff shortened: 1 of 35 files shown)


Engines/Wine/Verbs/DXVK/script.js
```diff
@@ -58,3 +59,35 @@ Wine.prototype.DXVK = function (dxvkVersion) {
 
     return this;
 }
+
+/**
+ * Verb to install DXVK
+*/
+var verbImplementation = {
```
madoar: I would like some comments or paragraphs in this verb implementation which describe/section what happens when

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -1,59 +1,80 @@
...
+        .set("native", ["d3d9"])
+        .do();
+
+    return this;
+}
+
+/**
+ * Verb to install VK9
+*/
+var verbImplementation = {
```
madoar: See above

Engines/Wine/Verbs/Windows XP SP 3/script.js
```diff
@@ -38,3 +39,22 @@ Wine.prototype.sp3extract = function (fileToExtract) {
 
     return this;
 };
+
+/**
+ * Verb to install Windows XP Service Pack 3
+*/
+var verbImplementation = {
```
madoar: See above

Engines/Wine/Verbs/DXVK/script.js
plata: I don't quite understand. It shows the wizard to select the version and then installs the selected version. What comments would you like to have

madoar: I would wish for something like:
```javascript 
/**
 * 1. download latest dxvk
 * 2. select the to be installed version
 **/
```

In addition I must confess that I have no clue about what version we're talking, because the download already happened?

Engines/Wine/Verbs/Windows XP SP 3/script.js
plata: I'm not so sure about the functionality of this verb myself. I commented to my best guess. In any case, I think this is pretty much obsolete (Windows XP...).

Engines/Wine/Verbs/VK9/script.js
plata: Done.

Engines/Wine/Verbs/DXVK/script.js
plata: Done.

PR #735, opened 2018-12-17
Add dotnet20

Zemogiter:
Help needed.
Despite copying what was done at winetricks the installer for dotnet20 throws this error:

zrzut ekranu z 2018-12-17 23-03-25


qparis:
When does the error occur?

Zemogiter:
@qparis at `this.run(setupFile, [setupFile, "/q:a", "/c:\"install.exe /q\""], null, false, true);`

qparis:
Try to omit the quotes in the third argument

Zemogiter:
No it doesn't.

qparis:
Same error?

qparis:
> No it doesn't.

qpmac:~ qparis$ cat test.py 
import sys

print(sys.argv)
qpmac:~ qparis$ python test.py /q /c:"install.exe /q"
['test.py', '/q', '/c:install.exe /q']

This should work pretty much like this

Engines/Wine/Verbs/dotnet20/script.js
```diff
@@ -0,0 +1,43 @@
...
+include(["engines", "wine", "verbs", "luna"]);
+include(["utils", "functions", "filesystem", "files"]);
+include(["engines", "wine", "plugins", "windows_version"]);
+
+/**
+* Verb to install .NET 2.0
+* @returns {Wine} Wine object
+*/
+Wine.prototype.dotnet20 = function () {
+    if (this.architecture() = "x86") {
```
qparis: Be careful, you forgot the == (instead of a simple =)

Engines/Wine/Verbs/dotnet20/script.js
```diff
@@ -0,0 +1,43 @@
...
+            .name("NetFx64.exe")
+            .get();
+    }
+
+    this.uninstall("Mono");
+
+    this.wizard().wait(tr("Please wait while {0} is installed ...", ".NET Framework 2.0"));
+    this.run(setupFile, [setupFile, "/q:a", "/c:install.exe /q"], null, false, true);
+
+    if (this.architecture() = "x86"){
```
qparis: Same here

Zemogiter:
Now the errors are gone aside from the on from screenshot.

qparis:
What if you try:
    this.run(setupFile, [setupFile, "/q", "/c:install.exe /q"], null, false, true);

qparis:
Ok, I understand the problem now.

wine.run("dotnetfx.exe", ["/q:a", "/c:install.exe /q"], null, false, true)

instead of 

wine.run("dotnetfx.exe", ["dotnetfx.exe", "/q:a", "/c:install.exe /q"], null, false, true)

Engines/Wine/Verbs/dotnet20/script.js
```diff
@@ -0,0 +1,43 @@
...
+            .get();
+    }
+
+    this.uninstall("Mono");
+
+    this.wizard().wait(tr("Please wait while {0} is installed ...", ".NET Framework 2.0"));
+    this.run(setupFile, [setupFile, "/q:a", "/c:install.exe /q"], null, false, true);
+
+    if (this.architecture() == "x86"){
+        remove(this.system32directory() + "/msvc" + "?" + "80.dll");
```
qparis: I'm afraid that this won't work. We need to enumerate all the files

Zemogiter:
Just tested this verb in my Sims 3 script. The game now launches. Verb is fully functional.

Issue #751, opened 2018-12-23
Crowdin link on README.md


ovari:
Can a link to the Crowdin be added to the README.md file at:
https://github.com/PhoenicisOrg/scripts/

It probably relates to:
https://github.com/PhoenicisOrg/scripts/tree/master/i18n

It could be similar to the link at:
https://github.com/PhoenicisOrg/phoenicis#translate

What do you think?

Thank you

madoar:
Strictly speaking there is a crowdin badge at the top of the README.
If you click on it you are forwarded to the crowdin page for the scripts.

At the same time I don't see a reason not to add an additional link to the README page, which makes it clear where the translations can be found.

ovari:
Thanks for updating @plata. Should there be another sentence like:

> If your language is not listed, please create an [issue](https://github.com/PhoenicisOrg/scripts/issues).

What do you think?

plata:
Makes sense. Can you create a PR for both repos?

ovari:
https://github.com/PhoenicisOrg/phoenicis/pull/1773
https://github.com/PhoenicisOrg/scripts/pull/817

PR #752, opened 2018-12-23
Little fixes (update README.md, ellipses, 32-bit and 64-bit, smart quotation marks)

ovari:
fixes https://github.com/PhoenicisOrg/scripts/issues/751


ovari:
Sorry, don't know how to do multiple separate pull requests using the GitHub's web based GUI.

plata:
You have to create multiple branches in your fork.

ovari:
@plata sorry we can't

ovari:
Feel free to change any files how you see fit.

plata:
The problem is: Some of the things you changed (e.g. the app descriptions) are just quoted from the official description. If you change them now because you prefer different quotation marks etc, it's only causing trouble (for very little benefit).

ovari:
32-bit and 64-bit
https://en.wikipedia.org/wiki/32-bit
https://en.wikipedia.org/wiki/64-bit_computing

Why is a thesis required to be written for little changes?

> just quoted from the official description

Some official descriptions have smart quotes, some don't. So POL5 will have a combination. It was understood that you didn't want a combination. Now it is understood that it is okay. Where is the consistency?

ovari:
We had updated all the POL5 files so that they were consistent with common best practice. 64-bit has been updated in another place and the PR accepted by @qparis without requiring a thesis to be written. Writing so much means less time for improving your, and other peoples’, software projects. We thought @plata you would appreciate this PR as we previous had helped you update the English strings to be better (concise, understandable, etc). Please accept our apologies that we unable to help your project to the best of our ability as we are discouraged by your process.

qparis:
@ovari I know that you may be frustrated but please do not take this personally, it does not worth it. 

I think you have misunderstood @plata because to my comprehension, we have never denied your merge request, so you should not have deleted it because it would probably have been accepted. 

In POLv5, we just want to create a culture of share and discussion about every change that is made. And this discussion can takes some time, generate some frustration, but will surely lead to a better product. If you want to go fast, go alone, if you want to go far, go together. 

I hope that you understand that we are challenging every merge request, not because of these contents, but just because we want the best for the product.

plata:
@qparis thanks. I just wanted to point this out as well.

My concern was only to have different changes in one pull request because it will make it hard for us to track which decisions were made (and more importantly why). However, exactly this point is crucial if we want to achieve a common style (which is the goal of all of us, including you @ovari).

PR #755, opened 2018-12-28
Update The Witcher 3: Wild Hunt, replaced GLSL with DXVK

Zemogiter:
After some testing I've concluded the framerate is better with vulkan while not suffering from ocasional graphical glitches from opengl.


plata:
This implies that some users might not be able to use the script anymore, right? I'm just thinking if it makes sense to provide both options.

ImperatorS79:
Note that all modern computer should have vulkan. Maybe a command to check if the computer has vulkan support can be useful. However, that implies the user has installed the driver (which is not necessary the case).

ImperatorS79:
It's possible, but generally GPU which does not have vulkan support will give bad performance. You can also ask the user if he wants to install dxvk (a way to revert it later could be interesting in this case).

ImperatorS79:
It's possible, but generally GPU which does not have vulkan support will give bad performance. You can also ask the user if he wants to install dxvk (a way to revert it later could be interesting in this case)

PR #757, opened 2018-12-29
Update dotnet452, removed check for x32 architecture

Zemogiter:
That if block was breaking `dotnet461` installation thus breaking the `Space Engineers` script.


plata:
I wonder why this has been in originally. I mean there must have been a reason. Maybe winetricks checks this as well? If so, why?

Zemogiter:
Winetricks gives warning but dosen't stop the installation of `dotnet452`.

Engines/Wine/Verbs/dotnet452/script.js
```diff
@@ -69,5 +66,4 @@ var verbImplementation = {
 };
 
 /* exported Verb */
-var Verb = Java.extend(org.phoenicis.engines.Verb, verbImplementation);
-
+var Verb = Java.extend(org.phoenicis.engines.Verb, verbImplementation);
```
plata: Please add newline add EOF.

PR #759, opened 2018-12-30
Remove trailing whitespace

plata:


plata:
@ovari did I forget it anywhere?

madoar:
The changes in the stream script look strange to me, because basically the whole file is modified?

plata:
I have no idea why it's shown like this.

plata:
@madoar please check again. This is only a GithHub diff issue. If you diff it locally, it shows only one line.

PR #765, opened 2019-01-01
Fix dos support issue

qparis:


qparis:
I think the error does not come from my change

qparis:
@plata Can we merge this commit in alpha-2 branch?

PR #768, opened 2019-01-01
Add GOG Galaxy

seijikun:
Hey guys. Awesome job with Alpha 2 :+1: 

GOG is quite moody and needs a weird symlink in the wine prefix.
But with that, everything seems to work fine.
(Except for errors after a game-installation finishes - which can be
ignored)


Applications/Games/GOG Galaxy/Online/script.js
```diff
@@ -0,0 +1,37 @@
...
+            .applicationHomepage("https://www.gog.com/")
+            .author("Markus Ebner")
+            .url("https://content-system.gog.com/open_link/download?path=/open/galaxy/client/setup_galaxy_1.2.51.30.exe")
+            .checksum("faa29721de18a171561a3e4dfa243953ff98aa09")
+            .executable("GalaxyClient.exe")
+            .wineArchitecture("amd64")
+            .wineVersion(LATEST_STAGING_VERSION)
+            .wineDistribution("staging")
+            .preInstall(function (wine/*, wizard*/) {
+                wine.windowsVersion("win10");
```
ImperatorS79: Unless it's needed to run the application, do not use it. Wine defaults to windows 7, and it should work for most apps.

seijikun:
Works without explicitly stating a windowsVersion, so I removed it.

plata:
@seijikun do you think it would make sense to install GOG Galaxy in the GOG QuickScript?

qparis:
Why not after all, we are supporting Steam, UPlay, and others

plata:
Could there be any reason not to have it? I mean it's optional so maybe there are people who do not want it?

qparis:
I think this script does not force you to have it, does it?

plata:
No but if we add it to the GOG QuickScript it will be installed for all GOG games.

qparis:
Sorry, I misunderstood what you said. If we add it to GoGScript, we should let the choice to the user

plata:
So a choice in the wizard "Do you want to install GoG Galaxy?". Should work.

qparis:
Better, a menu with several install methods (Download directly, Gog Galaxy, Local setup file)

plata:
see #769

PR #770, opened 2019-01-01
Downloader improvements

qparis:


Engines/Wine/Plugins/DOS support/script.js
```diff
@@ -135,4 +135,6 @@ Wine.prototype.dosbox = function () {
             return this;
         }
     }
+
+    return this;
```
plata: Didn't we have this already?

Utils/Functions/Net/Download/script.js
```diff
@@ -120,7 +120,7 @@ Downloader.prototype.get = function () {
     }
 
     if (this._localFile) {
-        this._downloader.get(this._url, this._localFile, this._headers, function (progressEntity) {
+        var downloadFile = this._downloader.get(this._url, this._localFile, this._headers, function (progressEntity) {
```
plata: What's the difference between `downloadFile` and `_localFile`?

qparis: _localFile is what you set
downloadFile is what you get

Engines/Wine/Plugins/DOS support/script.js
qparis: Sometimes, git behaves weird

Utils/Functions/Net/Download/script.js
plata: I don't understand. Doesn't it mean "please download to `_localFile`"?

qparis: _localFile can be a directory in fact. So if it is the case, you'll get extra information from the return

plata: Shouldn't we make the interface clearer? So either it has to be a file or a directory? I think otherwise it will only be confusing.

qparis: I don't think so. When you use "cp file destination", destination can be a file or a directory. If it is a directory, the new file's name is destination/file. If it is a file, the new file name is "destination". 
This component behaves exactly the same

Engines/Wine/Plugins/DOS support/script.js
plata: Not sure what you did but I think you just have to merge master to this branch.

Utils/Functions/Net/Download/script.js
plata: Ok but this should be documented.

plata: Ok but this should be documented.

qparis: Ok

Engines/Wine/Plugins/DOS support/script.js
qparis: My branch looks exactly like the other ones. This is really strange

plata: I don't know. If you look at the commits in this PR, it already doesn't look right. Normally you should see only what you did in this PR.

Utils/Functions/Net/Download/script.js
plata: Should be enough if you update the JSDoc for the downloader.

PR #775, opened 2019-01-01
Improved downloader

qparis:


qparis:
Can I merge? The problem with github pages does not come from me

PR #780, opened 2019-01-03
README: add description how to use git clone as local repository

plata:


README.md
```diff
@@ -3,7 +3,11 @@
 [![Crowdin Badge](https://d322cqt584bo4o.cloudfront.net/phoenicis-scripts/localized.svg)](https://crowdin.com/project/phoenicis-scripts)
 
 # How to add a script
-1. Select the right category
+1. Clone this repository
+
+2. Add the local checkout as local repository in Phoenicis (see [instructions](https://phoenicisorg.github.io/phoenicis/Users/repository/#local-directory))
```
madoar: `... as a local repository ...`

README.md
```diff
@@ -3,7 +3,11 @@
...
+
+3. Select the right category
```
madoar: How do you select the category?

README.md
```diff
@@ -3,7 +3,11 @@
 [![Crowdin Badge](https://d322cqt584bo4o.cloudfront.net/phoenicis-scripts/localized.svg)](https://crowdin.com/project/phoenicis-scripts)
 
 # How to add a script
-1. Select the right category
+1. Clone this repository
```
madoar: Maybe also add the github link, which should be cloned

README.md
plata: Well, where does the script belong to.

madoar: Yes I know that, but normally when I read or hear `select` I think about clicking on it, which is not what you mean here.
I think you want to say something like:
```
3. Select the right category through the use of its path in the next steps
```

plata: I get your point but I don't think this makes it clearer.

madoar:
@plata it seems like the indentation of the PR is not right

plata:
Why?

PR #781, opened 2019-01-03
Improve the best practices for the creation of custom scripts

madoar:
See https://github.com/PhoenicisOrg/phoenicis/issues/1711

### DIFF
--- a/docs/_docs/General/best-practices.md
+++ b/docs/_docs/General/best-practices.md
@@ -7,12 +7,21 @@ toc: false
 
 This page describes best practices for writing your scripts.
 
-1. Fork the repository
-2. Create a new branch for your script
+When writing your own script you'll need to test it. To test a script Phoenicis doesn't only require the script, but also some context information. For example a miniature image used in the library after the script has been installed. Some other context information is extracted by Phoenicis via the directory structure, the script is located in.
+
+Therefore to test a script it needs to be embedded inside a repository. For this a local repository, which is based on the official script repository, is recommended. To setup your own local repository proceed the following steps:
+
+1. Fork the official scripts repository
+2. Create a new branch for your script on your fork
 3. Clone your fork to your local scripts repository (usually `~/.Phoenicis/repository/`)
 4. Checkout your branch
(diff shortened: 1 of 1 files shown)


docs/_docs/General/best-practices.md
```diff
@@ -7,12 +7,21 @@ toc: false
 
 This page describes best practices for writing your scripts.
 
-1. Fork the repository
-2. Create a new branch for your script
+When writing your own script you'll need to test it. To test a script Phoenicis doesn't only require the script, but also some context information. For example a miniature image used in the library after the script has been installed. Some other context information is extracted by Phoenicis via the directory structure, the script is located in.
+
+Therefore to test a script it needs to be embedded inside a repository. For this a local repository, which is based on the official script repository, is recommended. To setup your own local repository proceed the following steps:
```
qparis: Maybe we could add an option to "run a directory script"

docs/_docs/General/best-practices.md
```diff
@@ -7,12 +7,21 @@ toc: false
...
+
+1. Fork the official scripts repository
```
plata: We should change the README to use a fork as well.

docs/_docs/General/best-practices.md
madoar: I'm not really sure a "run a directory script" option would work.

Currently we require for the script to be correctly displayed in the applications tab of Phoenicis, that the script is contained inside a category directory. (@plata I can't find this in the script repository documentation?)
I know this is only for the representation in the applications tab, which is not really required to when directly executing a script, but I'm not sure which other information is extracted from the directory structure during script execution.
I think @plata has more knowledge about this than I.

docs/_docs/General/best-practices.md
madoar: Yes, you're right

docs/_docs/General/best-practices.md
plata: The scripts require information about their location in the repository. This is injected from Java (otherwise you would have to set it for every single script). We have an open issue to change this behavior but until then, it's not possible. To be honest, I also don't see the point of doing this: If you want to provide a script, you must put it in git anyways.

PR #786, opened 2019-01-04
README: use a fork

plata:


madoar:
Can you also fix the indentation issue with the enumeration? 
You haven't used enough spaces, for the categories in step 3 and all the content starting from step 5.

Issue #792, opened 2019-01-09
BUG: Battle.net applications fails on <GAMENAME>.exe


Kreyren:
Battle.net installers fails with unexpected error resulting in installing wineprefix, but not listing it in phoenicis's Library.

### STEPS TO REPRODUCE
Install any battle.net application without `.executable(Battle.net.exe)`. Log-in -> Error occurs prior to log-in.

### WORKAROUND
using `.executable(Battle.net.exe)` fixes the issue, but launches the Battle.net instead of the game -> Expected compatibility issues with end-users installing games on battle.net due missing dependencies in wineprefix that is optimized for other title.

EDIT: same error appears if the Battle.net installer is already installed.

ImperatorS79:
All the Battle.net installers ?

Kreyren:
@ImperatorS79 Correct, testing your hearthstone atm.

ImperatorS79:
@Keyren ok I think the error could come from the fact that you still have prefixes from old tentative of installing those games. Could you go in the containers tab and delete all containers without shortcut ?

Kreyren:
Afaik phoenicis checks if the executable was called for some reason.. That seems like horrible idea for applications that calls external installer/launcher.

Recommends removing this check or adapting a function for extermal launcher/installer if needed.

ImperatorS79:
No, this is just the function that create the shortcut, which is called when the installer ends. But yeah, this does not work is the installer call another .exe to install the game.

Kreyren:
@ImperatorS79 noted.

Zemogiter:
I have no issues with either Battle.net app and Starcraft 2. Check the scripts for both.

plata:
@Zemogiter would you mind taking the lead here and help @Kreyren figuring out why it doesn't work for him?

Kreyren:
@Zemogiter More info? I don't see how is my script different from `star craft II` will test

Zemogiter:
@Kreyren What wine version and what branch (upstream,staging,proton,dos_support,cx) and what architecture (x86 or x64) are you using and what game are you trying to install?

Kreyren:
@Zemogiter How is WINE version relevant since script is using precompiled WINE? Try removing starcraft wineprefix and use battle.net with mobile authentification.

plata:
@Zemogiter @Kreyren what's the state of this issue?

Kreyren:
iirc it's fixed.

PR #793, opened 2019-01-09
Games/Hearthstone : Update, fixes

Kreyren:
Bug : #792
Changes WINE environment from winxp to win7 -> winxp outputs error about unsupported OS.
Changes executable to workaround #792
Tested, working
Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

### DIFF
--- a/Applications/Games/Hearthstone/Online/script.js
+++ b/Applications/Games/Hearthstone/Online/script.js
@@ -1,11 +1,6 @@
 include(["engines", "wine", "quick_script", "online_installer_script"]);
-include(["engines", "wine", "plugins", "windows_version"]);
 include(["engines", "wine", "verbs", "vcrun2015"]);
 include(["engines", "wine", "verbs", "corefonts"]);
-
-// CHANGELOG
-// 2019/01/09 // Kreyren - Changed executable on Battle.net.exe, mandatory for battle.net games
-// 2019/01/09 // Kreyren - Changed WINE from winxp on win7 since winxp will soon be unsupported
 
 // BUG: https://github.com/PhoenicisOrg/scripts/issues/792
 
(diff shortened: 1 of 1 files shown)


ImperatorS79:
Please remove the changelog and the win7 line, because win7 is the default in wine 3.0+.

Kreyren:
@ImperatorS79 Agreed with the changelog, 

unsure about unseting win7, but seems sane

Applications/Games/Hearthstone/Online/script.js
```diff
@@ -1,22 +1,24 @@
 include(["engines", "wine", "quick_script", "online_installer_script"]);
-include(["engines", "wine", "plugins", "windows_version"]);
 include(["engines", "wine", "verbs", "vcrun2015"]);
 include(["engines", "wine", "verbs", "corefonts"]);
 
+// BUG: https://github.com/PhoenicisOrg/scripts/issues/792
+
+// TODO: OSX support, needs verification
```
qparis: Please do not add these kind of comments on the code

Zemogiter: I've found similar kind of comment in the `Europa Universalis 2` script.

ImperatorS79: Yep, I think I will just add a link to the website where the patches can be downloaded ^^.

Zemogiter: No there is a website with static links to those patches but I do not own that game. Anyway the reason I've pointed that out was because I was wondering if the comments in that script should also be removed.

Kreyren: Using comments in installers seems sane assuming they are in testing. -> Helps with script development. 

I would agree with not using them in non-testing scripts.

Kreyren:
Noticed that i'm pushing from master, fixing

PR #790, opened 2019-01-05
Verbs/Adobeair : Initial commit

Kreyren:
Initial commit

Required for League Of Legends's Launcher

TODO: make sure that packaging does not break https://www.adobe.com/legal/terms.html 6.2

Warning: c04068e and 0d1aba2 are not to be trusted, commited based on my experince with adobeair, winetricks and what i remember from javascript + i'm not familiar with functions of phoenicis -> Hoped to provide a foundation to make commit easier to finish.

PR #801, opened 2019-01-10
Add Adobeair

qparis:
Follow up #790

### DIFF
--- a/Engines/Wine/Verbs/adobeair/script.json
+++ b/Engines/Wine/Verbs/adobeair/script.json
@@ -1,0 +1,11 @@
+{
+  "scriptName"                 : "adobeair",
+  "id"                         : "adobeair",
+  "compatibleOperatingSystems" : [
+    "MACOSX",
+    "LINUX"
+  ],
+  "testingOperatingSystems" : [],
+  "free"                    : true,
+  "requiresPatch"           : false
+}
--- a/Engines/Wine/Verbs/adobeair/script.js
+++ b/Engines/Wine/Verbs/adobeair/script.js
@@ -1,0 +1,36 @@
+include(["engines", "wine", "engine", "object"]);
+include(["utils", "functions", "net", "resource"]);
+
+/**
+ * Verb to install adobeair
+ * @returns {Wine} Wine object
+ */
+Wine.prototype.adobeair = function () {
+    var adobeair = new Resource()
+        .wizard(this.wizard())
+        .url("https://airdownload.adobe.com/air/win/download/latest/AdobeAIRInstaller.exe")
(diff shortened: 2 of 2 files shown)


Engines/Wine/Plugins/Windows version/script.js
```diff
@@ -15,18 +15,26 @@ Wine.prototype.windowsVersion = function (version, servicePack) {
...
-        var servicePackNumber = servicePack.replace("sp", "");
-        regeditFileContent += "[HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows NT\\CurrentVersion]\n";
-        regeditFileContent += "\"CSDVersion\"=\"Service Pack "+ servicePackNumber +"\"\n";
-        regeditFileContent += "[HKEY_LOCAL_MACHINE\\System\\CurrentControlSet\\Control\\Windows]\n";
-        regeditFileContent += "\"CSDVersion\"=dword:00000"+servicePackNumber+"00\n";
+    var regeditFileContent;
+    if (version == null) {
+        regeditFileContent =
+            "REGEDIT4\n" +
+            "\n" +
```
plata: This part is the same, so you could actually change the code to:
```Javascript
regeditFileContent =
    "REGEDIT4\n" +
    "\n" +
    "[-HKEY_CURRENT_USER\\Software\\Wine]";

if (version) {
...
```

Engines/Wine/Verbs/adobeair/script.json
```diff
@@ -0,0 +1,11 @@
+{
+    "scriptName"                 : "adobeair",
```
plata: name can be "Adobe AIR"

Engines/Wine/Plugins/Windows version/script.js
qparis: No it is not if you look well

Engines/Wine/Verbs/adobeair/script.json
qparis: Done

Engines/Wine/Plugins/Windows version/script.js
plata: Ah the "-" in front of HKEY is different. Didn't see that.

Still would prefer
```Javascript
if (version)
```
over
```Javascript
if (version == null)
```

qparis: Not sure because if version is undefined, then, we switch back to the getter. The I think in this particular case, we really need version to be explicitely null

plata: I see. It's not so nice from an API perspective.

PR #825, opened 2019-01-18
Update vulkanSDK

ImperatorS79:
The verb did not create the reg entries, but now it only create the entry for Wow6432node.

@Zemogiter could you check ?


ImperatorS79:
Ok, tested with a .reg. If using wine regedit myreg.reg -> does not work. But if I import int from regedit itself from the GUI -> works. 

Maybe a bug in wine so I will simply add a wizard asking the user to add the key manually for now.

ImperatorS79:
Ready to be merged if no errors.

Engines/Wine/Verbs/vulkanSDK/script.js
```diff
@@ -20,18 +22,48 @@ Wine.prototype.vulkanSDK = function () {
...
+    
+    var message = "Please add the following keys to the regsitry after the instalation:\n\n"    +
+                  "[HKEY_LOCAL_MACHINE\\Software\\Khronos\\Vulkan\\Drivers]\n"                  +
+                  "\"C:\\\\Windows\\\\winevulkan.json\"=dword:00000000"                         +
+                  "\n\n"                                                            	        +
+                  "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Khronos\\Vulkan\\Drivers\\]n"    +
+                  "\"C:\\\\Windows\\\\winevulkan.json\"=dword:00000000" ;
 
-    return this;
+    this.wizard().message(message);
```
qparis: I don't understand why showing this to the user

qparis: Why it does not work? 
This is not acceptable to ask him to do it in terms of user experience

ImperatorS79: Seems my comment did not get uploaded ^^.

qparis: I've just seen it. This could not be a bug in wine

ImperatorS79: I have just tested it and it is.

qparis: The file must not be correctly formatted. Why would it not work just with this particular reg file?

qparis: Look, you've left a trailing "n" character

ImperatorS79:
Just copy paste this in a .reg
```
REGEDIT4

[HKEY_LOCAL_MACHINE\Software\Khronos\Vulkan\Drivers]
"C:\\Windows\\winevulkan.json"=dword:00000000
```
You can interchange`REGEDIT 4` and `Windows Registry Editor Version 5.00`

qparis:
Ok, and what is the key that works correctly?

ImperatorS79:
```
REGEDIT4

[HKEY_LOCAL_MACHINE\Software\Wow6432Node\Khronos\Vulkan\Drivers]
"C:\\Windows\\winevulkan.json"=dword:00000000
```

qparis:
Both work for me

ImperatorS79:
With `wine regedit hello.reg` ?

qparis:
Yes (except that the file name is not hello)

ImperatorS79:
Really strange, it does not work for me with wine 4.0-rc4 staging 64 bits.

qparis:
Ensure that the line returns are the same in both files

Engines/Wine/Verbs/vulkanSDK/script.js
```diff
@@ -1,52 +1,75 @@
...
+        //checksum changes everytime a new version is released
+        .name("vulkan-sdk.exe")
+        .get();
+
+    this.run(setupFile, "/S");
+    this.wait();
+
+    var pathVulkanJSON = this.prefixDirectory() + "drive_c/windows/winevulkan.json";
+    touch(pathVulkanJSON);
+    var contentVulkanJSON = '{\n'                                                                  +
```
qparis: Can you use JSON.stringify?

ImperatorS79: What is that ?

qparis: Instead of storing the JSON in a string variable, you could use 

var contentVulkanJSON = json.stringify({
"key": "value", 
... 
})

qparis: JSON.stringify *

Zemogiter:
I've tried the old script for `VulkanSDK` and it only created the registry entry for Wow6432Node. I'm going to test your commit then mine. I was sure it was working when I created the PR.

ImperatorS79:
The two errors are strange since those identifiers in the json have to be like this.

plata:
It's ok. You may ignore them.

plata:
@qparis do you want to do the review (you checked also before)?

ImperatorS79:
I think this is ready to be merged.

Zemogiter:
Just tested that and everything works expect for 32bit registry entry creation.

plata:
@Zemogiter so the 32bit registry issue should be fixed before merging this?

Zemogiter:
Yes

plata:
@ImperatorS79 can you do that?

ImperatorS79:
I have no idea where this issue. Nevertheless, the code was not working at all before this PR, so I suggest to merge this and open a new issue.

Zemogiter:
OK so here's the thing. I've checked on winetricks source code and saw they did not include `"\` in
```
"\"C:\\\\Windows\\\\winevulkan.json\"=dword:00000000" ;
```
After I've removed that and installed in osu wineprefix the `Drivers` key inside `Vulkan` is created which didn't happen before. `Drivers` is empty though but this confirms there might be something wrong with the script despite the 64bit section working.

ImperatorS79:
This is a \" instead, which just allows to type " in a string delimited by ... "

plata:
@ImperatorS79 please merge master to resolve the conflict.

ImperatorS79:
I will open an issue for the registry key as soon as this is merged ?

plata:
@ImperatorS79 don't forget to add some details to the merge description (not title) when you merge this.

PR #833, opened 2019-01-19
Update Wine LATEST_DEVELOPMENT_VERSION and LATEST_STAGING_VERSION to 4.0-rc7

plata:


madoar:
What about dosbox?

qparis:
Not sure we want to update dosbox distribution as often. Let's keep stable releases for this one

PR #838, opened 2019-01-21
Update VK9

ImperatorS79:
New version + version before 0.28.0 will not work with this script (not really useful version since VK9 is still in "alpha")

### DIFF
--- a/Engines/Wine/Verbs/VK9/script.js
+++ b/Engines/Wine/Verbs/VK9/script.js
@@ -68,7 +68,7 @@ var verbImplementation = {
         wine.prefix(container);
         var wizard = SetupWizard(InstallationType.VERBS, "VK9", java.util.Optional.empty());
         // query desired version (default: 0.28.1)
-        var versions = ["0.29.0", "0.28.1", "0.28.0"]; //this script is not able to install older version (VK9.conf mandatory)
+        var versions = ["0.29.0", "0.28.1", "0.28.0"]; //this script is not able to install older versions (VK9.conf mandatory)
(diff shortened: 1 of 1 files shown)


Engines/Wine/Verbs/VK9/script.js
```diff
@@ -68,7 +68,7 @@ var verbImplementation = {
         wine.prefix(container);
         var wizard = SetupWizard(InstallationType.VERBS, "VK9", java.util.Optional.empty());
         // query desired version (default: 0.28.1)
-        var versions = ["0.28.1", "0.28.0", "0.27.0", "0.26.0", "0.25.0", "0.23.0", "0.17.0", "0.16.0", "0.6.0", "0.4.0"];
+        var versions = ["0.29.0", "0.28.1", "0.28.0"];
```
plata: Add comment why the older versions cannot be selected (just to avoid that someone adds them again later because he thinks they are missing).

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -68,7 +68,7 @@ var verbImplementation = {
         wine.prefix(container);
         var wizard = SetupWizard(InstallationType.VERBS, "VK9", java.util.Optional.empty());
         // query desired version (default: 0.28.1)
-        var versions = ["0.29.0", "0.28.1", "0.28.0"];
+        var versions = ["0.29.0", "0.28.1", "0.28.0"]; //this script is not able to install older version (VK9.conf mandatory)
```
plata: "older versions"

PR #839, opened 2019-01-22
Update dotnet (add versions +fix)

### LINKED ISSUE #705: osu! errors out when installing
I'm confused if this is a problem with Wine, or with POL, or even with my system,
When I install osu! normally through POL5, this shows up: 
https://imgur.com/a/zr3PZxA
PB: https://pastebin.com/9BRYEnai
This happens system-wide so I am a bit confused why it doesn't work in the first place.

System: Arch x86_64

ImperatorS79:
The modification are mainly to be closer to winetricks source (since those verbs come from there).

Added dotnet472, which will allow osu to install and fix #705 (upcoming PR), and also maybe paint.net.

Edit: [https://github.com/Winetricks/winetricks/issues/971](https://github.com/Winetricks/winetricks/issues/971) This could be interesting to implement, for example Space engineers.


Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -3,12 +3,21 @@ include(["engines", "wine", "plugins", "override_dll"]);
...
 * Verb to install .NET 4.0
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet40 = function () {
+    if (this.architecture = "amd64") {
+        print(tr("This package ({0}) may not fully work on a 64-bit installation. 32-bit prefixes may work better.", "dotnet40"));
+    }
+
+    var OSVersion = this.windowsVersion();
+    if (OSVersion == null)
```
plata: Can this actually happen?

Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -3,12 +3,21 @@ include(["engines", "wine", "plugins", "override_dll"]);
...
 /**
 * Verb to install .NET 4.0
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet40 = function () {
+    if (this.architecture = "amd64") {
+        print(tr("This package ({0}) may not fully work on a 64-bit installation. 32-bit prefixes may work better.", "dotnet40"));
+    }
+
+    var OSVersion = this.windowsVersion();
```
plata: Should be `osVersion`.

Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -30,14 +36,22 @@ Wine.prototype.dotnet40 = function () {
     this.wizard().wait(tr("Please wait while {0} is installed...", ".NET Framework 4.0"));
     this.run(setupFile, [setupFile, "/q", "/c:\"install.exe /q\""], null, false, true);
 
+    this.wizard().wait(tr("Please wait ..."));
```
plata: Should be "Please wait..." (no space before ellipsis). Also below.

Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -30,14 +36,22 @@ Wine.prototype.dotnet40 = function () {
...
     this.overrideDLL()
         .set("native", ["mscoree"])
         .do();
 
-    this.wizard().wait(tr("Please wait..."));
-    this.run("reg", ["add", "HKLM\Software\Microsoft\NET Framework Setup\NDP\v4\Full", "/v", "Install", "/t", "REG_DWORD", "/d", "0001", "/f"], null, false, true);
-    this.wizard().wait(tr("Please wait..."));
-    this.run("reg", ["add", "HKLM\Software\Microsoft\NET Framework Setup\NDP\v4\Full", "/v", "Version", "/t", "REG_SZ", "/d", "4.0.30319", "/f"], null, false, true);
+    this.wizard().wait(tr("Please wait ..."));
+    this.run("reg", ["add", "HKLM\\Software\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full", "/v", "Install", "/t", "REG_DWORD", "/d", "0001", "/f"], null, false, true);
```
plata: Can't you use the regedit plugin for the registry operations you need?

Engines/Wine/Verbs/dotnet46/script.json
```diff
@@ -0,0 +1,14 @@
+{
+    "scriptName"                 : ".NET 4.6",
+    "id"                         : "dotnet46",
+    "compatibleOperatingSystems" : [
+        "MACOSX",
+        "LINUX"
+    ],
+    "testingOperatingSystems" : [
```
plata: Why testing?

Engines/Wine/Verbs/dotnet461/script.js
```diff
@@ -2,14 +2,21 @@ include(["engines", "wine", "engine", "object"]);
...
 include(["utils", "functions", "filesystem", "files"]);
-include(["engines", "wine", "verbs", "dotnet452"]);
+include(["engines", "wine", "verbs", "dotnet46"]);
 
 /**
 * Verb to install .NET 4.6.1
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet461 = function () {
+    if (this.architecture = "amd64") {
```
plata: `=`? I think you want `===` or at least `==`.

Engines/Wine/Verbs/dotnet40/script.js
ImperatorS79: Yes, when you execute the script in a preInstall without this condition, the os version set at the end of the script does not work because the version is null.

Zemogiter:
You must replace `=` with `==`.

Engines/Wine/Verbs/dotnet40/script.js
qparis: You want to use a windows windows only during the installation, correct? 
If so, you can ommit this check since a recent commit I have done. In fact, now, windowsVersion(null) will unset the setting

ImperatorS79:
@qparis I have the same problem than in #825 with the registry: from the GUI of regedit with a .reg it works, with wine by command line or in the script it does not :thinking:.

Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -30,14 +33,28 @@ Wine.prototype.dotnet40 = function () {
...
-    this.wizard().wait(tr("Please wait..."));
-    this.run("reg", ["add", "HKLM\Software\Microsoft\NET Framework Setup\NDP\v4\Full", "/v", "Version", "/t", "REG_SZ", "/d", "4.0.30319", "/f"], null, false, true);
+    var regeditFileContent = "REGEDIT4\n"                                                                      +
+                             "\n"                                                                              +
+                             "[HKEY_LOCAL_MACHINE\\Software\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full]\n" +
+                             "\"Install\"=dword:0001\n"                                                        +
+                             "\"Version\"=\"4.0.30139\"";
+
+    this.regedit().patch(regeditFileContent);
+    print(regeditFileContent);
```
madoar: I guess this is a debug print?

Engines/Wine/Verbs/dotnet40/script.js
```diff
@@ -1,14 +1,20 @@
...
-include(["utils", "functions", "filesystem", "files"]);
+include(["engines", "wine", "plugins", "regedit"]);
+include(["engines", "wine", "plugins", "windows_version"]);
 
 /**
 * Verb to install .NET 4.0
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet40 = function () {
+    if (this.architecture == "amd64") {
```
madoar: Are you sure about `this.architecture` and not `this.architecture()`? `Wine.prototype.architecture` is a function, see:
https://github.com/PhoenicisOrg/scripts/blob/f7c70527bc69dab5f0aa8bef49fda2674d9e6575/Engines/Wine/Engine/Object/script.js#L29-L43

Engines/Wine/Verbs/dotnet45/script.js
```diff
@@ -2,21 +2,18 @@ include(["engines", "wine", "engine", "object"]);
...
 /**
 * Verb to install .NET 4.5
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet45 = function () {
-
-    if (this.architecture() == "amd64") {
-        throw "{0} cannot be installed in a 64bit wine prefix!".format("dotnet45");
+    if (this.architecture == "amd64") {
+        print(tr("This package ({0}) may not fully work on a 64-bit installation. 32-bit prefixes may work better.", "dotnet45"));
```
madoar: Why `print(...)` and not `this.wizard().message(...)`?

ImperatorS79: Well, I do not know if it is relevant to display this when you are installing an application. But sure it should be done when installing the verb from the container tab.

Engines/Wine/Verbs/dotnet40/script.js
madoar: This needs to be changed in the other files too.

Engines/Wine/Verbs/dotnet45/script.js
ImperatorS79: Yes, but I wonder if showing such a warning when installing an app will not confuse people like "this does not seem really stable". I mean the script should be tested so this warning is meaningless when the verb is called inside an application script.

madoar: Then why add the warning at all?

ImperatorS79: Just as a reminder, for people wanting to use the verb (I think I will add it as a message in the verb implementation).

madoar: @plata @qparis what do you think? How should we print warnings? Via `print(...)` or `this.wizard().message(...)`? I think we need to have a standard for this.

plata: Depends if the user can react on it. In this case, showing the message in the wizard would mean that a user installing a script which uses this verb would see the message. What is he supposed to do about it (except opening an issue on GitHub asking if this is alright)?

madoar: Then why do we show the message at all? Everything we print with `print` is only written in the terminal, which is not really a place where I would look for warning messages as a user.

Maybe we can add an icon with an `?` or an `!` which shows the warning when hovering over it.
This would still add the warning to the UI but hide it to prevent the user from overreacting?

plata: If it's in the terminal, we should get it as part of a potential error report (which is IMHO the only use case for this).

ImperatorS79:
I think I will end up only modifying dotnet40 and dotnet45, because all the others seems completely broken now (which is strange since I made dotnet472 works some times ago...).

Zemogiter:
In `Remove Mono` the id have a upper case

ImperatorS79:
I have seen ;-).

ImperatorS79:
@Zemogiter osu! should install without any problem. The bug came from the fact that, when @qparis modified ´wine.windowsVersion()` to delete `[HKEY_CURRENT_USER\\Software\\Wine]` when calling `wine.windowsVersion(null)`, this also deleted dll overrides information !

plata:
@qparis have you been aware of this effect? Is it what you intended/how it should behave?

qparis:
> @qparis have you been aware of this effect? Is it what you intended/how it should behave?

No, it is not normal. Can someone help to debug this?

ImperatorS79:
This simply because you delete `[HKEY_CURRENT_USER\Software\Wine]`, which thus deletes `[HKEY_CURRENT_USER\Software\Wine\DllOverrides]`.

qparis:
Ok, we should use instead: 
`[HKEY_LOCAL_MACHINE\Software\Test]
"testkey1"=-`

ImperatorS79:
Normally everything is ready here.

ImperatorS79:
Is it possible to check whether or not `this.uninstall("Mono")` succeed ? I mean, if mono is already uninstalled, We do not want to delete the registry key.

plata:
There`s a check in `Wine.prototype.uninstall` already. The only thing you need to do is change the return type of that method to boolean. After that, verify that nothing is concatenated after the `uninstall()` call in the existing scripts.

Engines/Wine/Verbs/Remove Mono/script.js
```diff
@@ -0,0 +1,45 @@
+include(["engines", "wine", "engine", "object"]);
+include(["utils", "functions", "filesystem", "files"]);
+include(["engines", "wine", "plugins", "regedit"]);
+
+/**
+* Verb to remove mono
+* @returns {Wine} Wine object
+*/
+Wine.prototype.removemono = function () {
```
madoar: I would recommend to name the method `removeMono` and its `id`: `remove_mono`

Engines/Wine/Verbs/dotnet20/script.js
```diff
@@ -1,43 +1,57 @@
...
 include(["engines", "wine", "plugins", "windows_version"]);
+include(["engines", "wine", "verbs", "removemono"]);
 
 /**
 * Verb to install .NET 2.0
 * @returns {Wine} Wine object
 */
 Wine.prototype.dotnet20 = function () {
+    var osVersion = this.windowsVersion();
+    if (osVersion == null)
```
madoar: I would recommend to add a `{` and `}` around the if-then body. This makes it easier to read and prevents errors.

Engines/Wine/Verbs/Remove Mono/script.js
ImperatorS79: an id cannot contain an _.

madoar: Are you sure? What about:

https://github.com/PhoenicisOrg/scripts/blob/92fe6100542f3919e09cfd0de61575cee542cda9/Engines/Wine/Verbs/dotnet20/script.js#L5
 
and

https://github.com/PhoenicisOrg/scripts/blob/92fe6100542f3919e09cfd0de61575cee542cda9/Engines/Wine/Plugins/Windows%20version/script.json#L3

ImperatorS79: OK, I remembered codacy complaining about a _ somewhere. So if it is not the id , it was certainly the method name ^^.

Zemogiter:
Please check codacy.

Engines/Wine/Verbs/Remove Mono/script.js
plata: We should have a documentation about this.

Engines/Wine/Verbs/dotnet20/script.js
plata: If Codacy doesn't complain about this, we should configure it to do so.

ImperatorS79:
If the checks are good, I do not personally see any more problems with this.

Engines/Wine/Verbs/dotnet20/script.js
```diff
@@ -1,43 +1,55 @@
...
-        var setupFile = new Resource()
+        var setupFile64 = new Resource()
             .wizard(this.wizard())
             .url("https://download.microsoft.com/download/a/3/f/a3f1bf98-18f3-4036-9b68-8e6de530ce0a/NetFx64.exe")
             .checksum("e59cca309463a5d98daeaada83d1b05fed5126c5")
             .name("NetFx64.exe")
             .get();
-        this.uninstall("Mono");
+
+        this.remove_mono();
```
madoar: Not `this.removeMono();`?

Zemogiter: Correct.

ImperatorS79:
I do not manage to fix the conflicts by following the command line instructions, so if anyone more experienced could try, that would be great.

plata:
Normally, if you merge master this should solve the conflicts.

ImperatorS79:
This would lead to conflicts.

plata:
Sure, you need to resolve the conflicts during the merge.

ImperatorS79:
And I do not know how to do it ^^.

plata:
https://git-scm.com/docs/git-merge

Basically:
```
git merge master
git mergetool -t meld
git commit
```

ImperatorS79:
Where ? In my local repo ? On the dotnet branch ? How to I keep what I have committed ?

Zemogiter:
What I do is just click on the "Resolve confilcts" button on the website and manually remove conflicts.

ImperatorS79:
This button is greyed because "conflicts are too complex" -_-'.

Zemogiter:
lol how could a couple of lines be too complex?

ImperatorS79:
¯\_(ツ)_/¯

Zemogiter:
I found this. Not sure if it helps though:
https://ashfurrow.com/blog/resolving-github-merge-conflicts/

plata:
@ImperatorS79 in your local clone with the branch you used to create the PR (you master branch needs to be up to date of course).

ImperatorS79:
Hoping nothing else must be changed ^^.

plata:
Some of the files changed from 644 to 755. Is that on purpose?

ImperatorS79:
Really ? I did not modified any permission on my pc.

plata:
At least the diff says so...

PR #845, opened 2019-01-25
Update Faudio verb

ImperatorS79:
Only call `this.system64directory()` for 64 bits container.


ImperatorS79:
No, I still do not know why @Zemogiter have an issue that really looks like the .tar.xz extraction issue.

PR #850, opened 2019-02-01
Update Subnautica, using latest stable Wine version and added a postInstall warning regarding shader mods.

Zemogiter:


Applications/Games/Subnautica/Steam/script.js
```diff
@@ -15,21 +15,24 @@ var installerImplementation = {
...
                 wine.DXVK();
-                wizard.message(tr("Please ensure you have the latest drivers (415.22 minimum for NVIDIA and mesa 19 for AMD) or else this game might not work correctly."));
-                wizard.message(tr("You can make the game smoother by using this: https://github.com/lutris/lutris/wiki/How-to:-Esync"));
                 var screenSize = java.awt.Toolkit.getDefaultToolkit().getScreenSize();
                 wine.setVirtualDesktop(screenSize.width, screenSize.height);
                 var registrySettings = new AppResource().application([TYPE_ID, CATEGORY_ID, APPLICATION_ID]).get("fix.reg");
                 wine.regedit().patch(registrySettings);
             })
+            .postInstall(function (wine, wizard) {
+                wizard.message(tr("Due to potential confilct with Vulkan, shader mods break the game (the executable file works but no window is displayed)."));
```
plata: "a potential conflict"

Applications/Games/Subnautica/Steam/script.js
```diff
@@ -15,21 +15,24 @@ var installerImplementation = {
             .author("Zemogiter")
             .applicationHomepage("https://unknownworlds.com/subnautica/")
             .wineDistribution("upstream")
-            .wineVersion("4.0-rc1")
+            .wineVersion("4.0")
```
plata: This is exactly what we wanted to avoid. Please use the latest stable version or latest development version.

Applications/Games/Subnautica/Steam/script.js
```diff
@@ -15,21 +15,24 @@ var installerImplementation = {
...
             .wineArchitecture("amd64")
             .appId(264710)
             .preInstall(function (wine, wizard) {
+                wizard.message(tr("Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else this game might not work correctly."));
+                wizard.message(tr("You can make the game smoother by using this: https://github.com/lutris/lutris/wiki/How-to:-Esync"));
                 wine.vcrun2013();
                 wine.corefonts();
                 wine.d3dx11();
                 wine.DXVK();
```
ImperatorS79: Do yo still needs d3d11 verb with DXVK (here it is working because the DXVK verb is after the d3d11)?

Applications/Games/Subnautica/Steam/script.js
Zemogiter: This dosen't happen when the script is ran without dxvk so I think it makes sense.

Applications/Games/Subnautica/Steam/script.js
Zemogiter: If we were to do that and the next stable/dev version of wine introduced some regression, we would receive a bunch of issue reports down the line that we won't be able to fix because it wouldn't be an issue on our end. This applies not just to Subnautica but any app in theory. By using specific versions of Wine we ensure the script will work not just for us but for everyone else.

Applications/Games/Subnautica/Steam/script.js
Zemogiter: I did not test it with just dxvk. I've tested it the other way around (d3dx11 without dxvk) when the game failed to display a window when using shader mods.

Applications/Games/Subnautica/Steam/script.js
ImperatorS79: Put LATEST_STABLE_VERSION and it will stay at 4.0 for a year ^^.

Applications/Games/Subnautica/Steam/script.js
ImperatorS79: Please test without d3d11.

Applications/Games/Subnautica/Steam/script.js
```diff
@@ -15,21 +14,23 @@ var installerImplementation = {
             .author("Zemogiter")
             .applicationHomepage("https://unknownworlds.com/subnautica/")
             .wineDistribution("upstream")
-            .wineVersion("4.0-rc1")
+            .wineVersion(LATEST_STABLE_VERSION)
             .wineArchitecture("amd64")
             .appId(264710)
             .preInstall(function (wine, wizard) {
+                wizard.message(tr("Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else this game might not work correctly."));
```
ImperatorS79: I wonder if this should simply be added to the dxvk verb.

Zemogiter: Makes sense.

ImperatorS79: Then maybe change it to be an actual message.

Zemogiter: Those print messages dosen't show up in script execution.

plata:
@ImperatorS79 I'll leave the review for you.

Issue #858, opened 2019-02-07
Script (under Test): Elite Dangerous + Steam


suamor:
I'm trying out the script to install EliteDangerous. Unfortunately I'm stuck in the Steam Installation step. 
Up to this step the installation seemed to work fine (wine 4.0 staging, .Net 4.5.2, DX10 and other dependencies installed, not sure why staging was used as stable is released since a longer time).

I have Elite Dangerous from Frontier Store. It is a quite common way to get Elite Dangerous.
Steam should be one of two options (select Installer Package or Download from Steam).

Also it would be great to know how to proceed from here (apart from looking into the script itself). 

System: Ubuntu 18.04 (with all updates installed).
Phonicis: SnapShot just downloaded and installed.

Zemogiter:
Just tested it and I got errors at `dotnet452` installation. The same I had in https://github.com/PhoenicisOrg/scripts/pull/840#issuecomment-460845318
Steam installed fine though.

suamor:
Yes, I have the same here. The Prefix was updated to 4.0 .
Steam works fine, I can install the ED Launcher too, but not start because .Net is broken.

Installing 4.0 via "Verbs" or any later versions fails, independently if wine-mono was installed previously or not. One of the scripts fails on this check and may crash or whatever (no log). 

I have not tried anything else than win10 as configuration option.

ImperatorS79:
> "mscorlib recursive resource lookup" and "Infinite recursion during resource lookup..." Resource Name: Arg_ExternalException.

Generally this is the sign of a broken .NET instalation.

> Up to this step the installation seemed to work fine (wine 4.0 staging, .Net 4.5.2, DX10 and other dependencies installed, not sure why staging was used as stable is released since a longer time).

You should not use the dx10 verb here, it is far better to use DXVK.

suamor:
Ok, then I will better try do a completely manual installation so I have a better control and understanding of the individual steps.

I found myself out that 4.5.2 is likely broken, because some winecfg entry  had not been set as described here for newer .Net installation including 4.5.2:
https://appdb.winehq.org/objectManager.php?sClass=version&iId=34219

The original 4.5.2 page states that additionally some registry entries need to be set manually. Not sure if this step is still required as it was not mentioned on the other page.

ImperatorS79:
[...]

PR #859, opened 2019-02-07
Update Elite: Dangerous, fix for #858

Zemogiter:
Since dotnet452, dotnet461 and dotnet462 (but oddly enough not dotnet472 which is used for `Space Engineers`) display the following errors: 
```
"mscorlib recursive resource lookup" and "Infinite recursion during resource lookup..." Resource Name: Arg_ExternalException.
```
regarding of which way you install those verbs (script, Verbs tab or winetricks) I've come with a solution. Basically I'm copying the latest AppDB entry except for dotnet version:
https://appdb.winehq.org/objectManager.php?sClass=version&iId=37490
@suamor can you test this new script? I do not own this game.


ImperatorS79:
I think you should keep quartz for the intro video. Why not use dotnet45 ?

Zemogiter:
@ImperatorS79 not sure since that AppDB didn't mention using quartz. That's why I asked @suamor
Also I can't test if dotnet45 works because when I'm clicking on Elite: Dangerous (or any other icon) in Containers tab the menu with Verbs tab as well as "Delete container" and "Open in file browser" buttons does not show up anymore. I'm on the latest repo and POL5.

ImperatorS79:
dotnet452 should work with #839.

Zemogiter:
@suamor can you test this now?

ImperatorS79:
@remove the 2,it is dotnet45 that works, sorry ^^.

Zemogiter:
Really? When I removed the prefix and ran the script with dotnet452 it worked. Can't test the game itself since I dont have it but the script works.

ImperatorS79:
It does not crash, but is does not install anything either.

Zemogiter:
I'm not sure but I browsed though registry and I saw nothing wrong. So maybe it will work?

ImperatorS79:
Believe me. The NDP-452-... .exe only displayed five fixme's when it is executed and then stops. There is clearly a problem.

Zemogiter:
Oh so you were testing it in POL4.

ImperatorS79:
No, in phoenicis.

Zemogiter:
I don't see anything in phoenicis.log

ImperatorS79:
I tested it using "Open a terminal" in a virgin container, making the steps myself. I also tested winetricks that way and it ended up the same.

Zemogiter:
If winetricks yielded the same result maybe it's a wine issue?

ImperatorS79:
That is what I think. But I wonder how something that important ended up in stable version.

Applications/Games/Elite:Dangerous/Steam/script.js
```diff
@@ -1,25 +1,23 @@
...
 var installerImplementation = {
     run: function () {
         new SteamScript()
             .name("Elite:Dangerous")
             .editor("Frontier Developments")
             .author("ImperatorS79")
-            .wineVersion(LATEST_STAGING_VERSION)
-            .wineDistribution("staging")
+            .wineVersion(LATEST_STABLE_VERSION)
+            .wineDistribution("upstream")
```
qparis: You can remove this as upstream is the default distribution. Unless it is really important to use upstream and you need to guarantee that it will be used.

plata:
@Zemogiter what's the state here? Shall it be reviewed?

Zemogiter:
The script is working but I need someone who owns the game to test it.

PR #864, opened 2019-02-11
Update Wine.uninstall: return a boolean

ImperatorS79:
Sorry for the commit, I did not see I was committing directly.


plata:
@ImperatorS79 do you still need the branch? Otherwise please delete it. In general, it is advisable to keep working in a fork even if you have write permission to the repository.

PR #865, opened 2019-02-13
Include by ID

### LINKED ISSUE #543: flexible include mechanism
Remove the fixed 4 level namespace and replace it by a flexible slug include mechanism.

How it could work:
The `ìd` in the `script.json` becomes a `slug`, e.g. `dotnet40` becomes `engines/wine/verbs/dotnet40`. The `include` changes from `include(["Engines", "Wine", "Verbs", "dotnet40"])` to `include("engines/wine/verbs/dotnet40")`.
A method `getAll` could return a list of `ScriptDTO` for a given slug, e.g. `getAll("applications/games")` would return `applications/games/aoeII/steam`, `applications/games/uplay/online` etc.

Pros:
* Removes the fixed 4 levels (e.g. strange paths like `Engines/Wine/Engine/Object`).
* Removes the artificial assignment of scripts to `ApplicationDTO` and `ScriptDTO` (why is "Verbs" an application?).
* Could allow to run a script via "Run a script" button (see PhoenicisOrg/POL-POM-5 #1179).

Cons:
* For application, the valid hierarchy of category/application/script is encoded in the slug instead of a class hierarchy. We could solve this by keeping the 4 level namespace for applications only but it's probably not a good idea to differentiate between scripts in the repository.

plata:
fixes #543 

requires https://github.com/PhoenicisOrg/phoenicis/pull/1833


plata:
I don't see ow this is related to this PR but why not.

madoar:
It is not really related to this PR, the idea just came to me while reviewing this PR, so I added it to the review ;)

PR #871, opened 2019-02-27
Update FAudio verb (new version)

ImperatorS79:


Zemogiter:
Shouldn't we make a list of faudio versions like we did with DXVK?

ImperatorS79:
Yes indeed

PR #874, opened 2019-03-01
Update DXVK: auto install dlls

ImperatorS79:
Old versions of DXVK does not have d3d10.


Engines/Wine/Verbs/DXVK/script.js
```diff
@@ -43,26 +43,30 @@ Wine.prototype.DXVK = function (dxvkVersion) {
...
+    var sys32dir = this.system32directory();
     var dxvkTmpDir = this.prefixDirectory() + "/TMP/dxvk-" + dxvkVersion;
+    var self = this;
 
-    cp(dxvkTmpDir + "/x32/d3d11.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/dxgi.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10core.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10_1.dll", this.system32directory());
+    forEach.call(ls(dxvkTmpDir + "/x32"), function (file) {
```
madoar: Please add a comment to the code explaining what you're doing in/with the loop.

Engines/Wine/Verbs/DXVK/script.js
```diff
@@ -43,26 +43,30 @@ Wine.prototype.DXVK = function (dxvkVersion) {
...
-    if (this.architecture() == "amd64") {
-        cp(dxvkTmpDir + "/x64/d3d11.dll", this.system64directory());
-        cp(dxvkTmpDir + "/x64/dxgi.dll", this.system64directory());
-        cp(dxvkTmpDir + "/x64/d3d10core.dll", this.system64directory());
-        cp(dxvkTmpDir + "/x64/d3d10.dll", this.system64directory());
-        cp(dxvkTmpDir + "/x64/d3d10_1.dll", this.system64directory());
+    if (this.architecture() == "amd64")
+    {
+        var sys64dir = this.system64directory();
+        forEach.call(ls(dxvkTmpDir + "/x64"), function (file) {
```
madoar: Please add a comment to the code explaining what you're doing in/with the loop.

Engines/Wine/Verbs/DXVK/script.js
madoar: Does `ls(dxvkTmpDir + "/x32").forEach(...)` not work?

ImperatorS79: This is the same as in Faudio verb

ImperatorS79: It works.

madoar: I still think it is a good idea to add a short comment describing the content and goal of the foreach loop. This really helps understanding the code :)

madoar: I think `ls(dxvkTmpDir + "/x32").forEach(...)` is more intuitive compared to `forEach.call(ls(dxvkTmpDir + "/x32"), ...)`

ImperatorS79: Ok, I think it does not work the other way ^^.

Engines/Wine/Verbs/DXVK/script.js
```diff
@@ -43,26 +43,32 @@ Wine.prototype.DXVK = function (dxvkVersion) {
...
-    cp(dxvkTmpDir + "/x32/dxgi.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10core.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10.dll", this.system32directory());
-    cp(dxvkTmpDir + "/x32/d3d10_1.dll", this.system32directory());
+    //Copy 32 bits dll to system* and apply override
+    forEach.call(ls(dxvkTmpDir + "/x32"), function (file) {
+        if (file.endsWith(".dll")) {
+            cp(dxvkTmpDir + "/x32/" + file, sys32dir);
+            self.overrideDLL()
+                .set("native", [file])
```
madoar: Should the `.dll` part of the filename be part of the override?

ImperatorS79: Not a problem since overrideDLL check for that.

PR #886, opened 2019-03-06
Use Java.type to access Java classes

plata:
see [Graal JavaScript to Java interoperability](https://github.com/graalvm/graaljs/blob/master/docs/user/JavaInterop.md)


plata:
That's a different topic. This PR doesn't really change anything. It only ensures that what we have currently works with Nashorn and Graal.

madoar:
Oh, I thought the filesystem module was usable with nashorn... 
Sorry about that

plata:
Now I'm confused.

madoar:
You don't need to be. I thought nashorn would support the filesystem module of JS, but this isn't the case because it doesn't seem to support the `require` command. Therefore my whole point is moot.

If we could use the filesystem module it would be the more clean solution with regards to https://github.com/PhoenicisOrg/phoenicis/issues/1630.

plata:
Understood.

PR #887, opened 2019-03-06
Secure Wine.regedit().patch()

plata:
If `patch()` is called from Javascript, the `patchContent` will normally be a simple string, i.e. the `getClass` method does not exist.


madoar:
By simple string you mean a JS string, right? In this case you should do comparison against "string" and not against `undefined`, right?

plata:
I'm testing if the function exists not if it's a string.

madoar:
function? Isn't `patchContent` either a string or a `byte[]`?

plata:
The check is
```Javascript
typeof patchContent.getClass
```
not
```
typeof patchContent
```

madoar:
Yes, but I don't understand why? An `undefined` check is normally used to check whether a parameter has been passed at all. A mixture between JavaScript `typeof` and Java `getClass` seems strange to me.

plata:
It verifies that the `getClass` function exists in `patchContent`. If it doesn't (e.g. because `patchContent` is a string), the code crashes (in Graal). In my opinion this is much better than checking if `patchContent` is a string because that's only an indirect proof (so it could be that it's not a string but still doesn't have a `getClass` function).

madoar:
My issue with your check is, that I expect it to lead to strange results if either no parameter is passed to the method, i.e. `patch()` is called, or if any kind of JavaScript object is directly passed, which shouldn't provide a `getClass` method.

plata:
Of course, if you pass completely invalid (or no) input, the function will not do anything useful. However, that's true for each and every function we have in Javascript. It's simply not statically typed. So the only thing this check does is to distinguish between two **valid** input types (even though I'm not sure if the byte[] part is used anywhere at all). Or do you think we should change all functions to first verify the input types?

madoar:
Hmmm, it seems like I read the method wrong. We're only accepting Java `byte[]` objects it seems.
But because a user could accidentally provide a JS object instead of a Java one we require the mentioned additional check.

plata:
With this PR, the graal engine settings issue is fixed so if you want to test the graalvm branch, feel free.

Issue #894, opened 2019-03-13
Request: Add OPSU!


Kreyren:
Requsting OPSU! to be added as alternative for OSU! (https://github.com/PhoenicisOrg/scripts/pull/840)

OPSU! is FOSS OSU!

Relevant: https://lutris.net/games/install/4903/view

Was reported to have issues with log-in

PR #897, opened 2019-03-14
Games/OPSU : Initial Commit

Kreyren:
Fixes : https://github.com/PhoenicisOrg/scripts/issues/894

HELP WANTED

Requires sanity check

### DIFF
--- a/Applications/Games/opsu/Online/script.js
+++ b/Applications/Games/opsu/Online/script.js
@@ -1,0 +1,20 @@
+include("engines.wine.quick_script.online_installer_script");
+
+var installerImplementation = {
+    run: function () {
+        new OnlineInstallerScript()
+            .name("osu!")
+            .editor("Jeffrey itdelatrisu Han")
+            .applicationHomepage("https://github.com/itdelatrisu/opsu")
+            .author("Kreyren")
+            .category("Games")
+            .executable("opsu-0.16.1.jar")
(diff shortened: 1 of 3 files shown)


ImperatorS79:
opsu! can run natively on linux, according do the homepage, so I will close this.

Kreyren:
> opsu! can run natively on linux, according do the homepage, so I will close this.

How is it relevant?

ImperatorS79:
Why would you want to do a wine script, for a .jar games that you can run natively on linux directly ?
You do not need wine in this case.

Kreyren:
@ImperatorS79 No need to use wine.

Since you need to download dependencies for lots of games that runs natively on linux -> Phoenicis can handle that better expecially when those games doesn't work on deps provided by distro like opsu on fedora that outputs:
```
OpenJDK 64-Bit Server VM warning: You have loaded library /home/jamespark/Games/osu/Natives/liblwjgl.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
```

Or other games that needs older version of deps that user don't neccesary want to install.

+ it would expand phoenici's usecase since games would be one click with required deps or methods like decising if you want to play windows version of OSU using WINE or native, etc..

Curent is that if you want to install a game -> you need to debug it.

PR #916, opened 2019-04-23
Games/Steam: Update wine

Kreyren:
Forcing latest staging since it's using 4.0 by default.

### DIFF
--- a/Applications/Games/Steam/Online/script.js
+++ b/Applications/Games/Steam/Online/script.js
@@ -10,6 +10,8 @@ var installerImplementation = {
             .url("https://steamcdn-a.akamaihd.net/client/installer/SteamSetup.exe")
             .checksum("4b1b85ec2499a4ce07c89609b256923a4fc479e5")
             .category("Games")
+            .wineDistribution("staging")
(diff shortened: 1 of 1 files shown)


Kreyren:
Results in
image

plata:
What's the purpose of this change?

Kreyren:
@plata Update wine for steam script assuming that it's using WINE 4.0.

plata:
I see that but why? In general, upstream should be preferred to staging.

Kreyren:
@plata

>  In general, upstream should be preferred to staging.

Based on what? Making something work on upstream is usually a miracle..

qparis:
> @plata
> 
> > In general, upstream should be preferred to staging.
> 
> Based on what? Making something work on upstream is usually a miracle..

Based on what? You should be more careful before claiming such assumptions. Wine is a perfectly stable software, and I agree with @plata. Staging is experimental and should be avoided when stable releases work.

Kreyren:
@qparis thats why i said usually a miracle.. fair enough then, closing.

PR #921, opened 2019-05-01
Add msxml3

Zemogiter:
Tested on `The Sims` wineprefix and it works.


Engines/Wine/Verbs/msxml3/script.js
```diff
@@ -0,0 +1,45 @@
...
+        .name("msxml3.msi")
+        .get();
+
+    remove(this.system32directory() + "/msxml3.dll")
+    this.overrideDLL()
+        .set("native", ["msxml3"])
+        .do();
+
+    this.wizard().wait(tr("Please wait while {0} is installed...", "msxml3"));
+    this.run(setupFile32, ["/q:a", "/c:msxml3.msi /q"], null, false, true);
```
ImperatorS79: This does not looks like winetricks one : w_try "$WINE" msiexec /i msxml3.msi /

Zemogiter: We don't have a separate command for msi files. But the installer works fine.

plata: Just for clarification: `wine.run` will behave correctly if it detects a msi (depending on the file suffix).

PR #938, opened 2019-05-13
Add dotnet20sp2

Zemogiter:
Needed for Sims 4 but getting this error:
[dotnet 20 sp2 error log.txt](https://github.com/PhoenicisOrg/scripts/files/3172804/dotnet.20.sp2.error.log.txt)
I've made the script similiar to winetricks so I don't know why winetricks works.


Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,76 @@
...
+    if (this.architecture() == "x86") {
+
+        var setupFile32 = new Resource()
+            .wizard(this.wizard())
+            .url("https://download.microsoft.com/download/c/6/e/c6e88215-0178-4c6c-b5f3-158ff77b1f38/NetFx20SP2_x86.exe")
+            .checksum("22d776d4d204863105a5db99e8b8888be23c61a7")
+            .name("NetFx20SP2_x86.exe")
+            .get();
+
+        this.removeMono();
```
ImperatorS79: You should remove mono at the beginning, not in each if elif else.

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,76 @@
...
+
+
+/**
+* Verb to install .NET 2.0 SP2
+* @returns {Wine} Wine object
+*/
+Wine.prototype.dotnet20sp2 = function () {
+    var osVersion = this.windowsVersion();
+    this.windowsVersion("winxp");
+    var dlls = [
```
ImperatorS79: On one line, and I think you should remove those dll overrides at the end (that is why winetricks use an env var).

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,76 @@
...
+            .wizard(this.wizard())
+            .url("https://download.microsoft.com/download/c/6/e/c6e88215-0178-4c6c-b5f3-158ff77b1f38/NetFx20SP2_x64.exe")
+            .checksum("a7cc6c6e5a4ad9cdf3df16a7d277eb09fec429b7")
+            .name("NetFx20SP2_x64.exe")
+            .get();
+
+        this.removeMono();
+
+        this.wizard().wait(tr("Please wait while {0} is installed...", ".NET Framework 2.0 SP2"));
+        this.run(setupFile64, ["/q:a", "/c:install.exe /q"], null, false, true)
```
ImperatorS79: This is not the command used in winetricks:
```
${W_OPT_UNATTENDED:+ /q /c:"install.exe /q"}
```
Where W_OPT_UNATTENDED is basically a boolean

Engines/Wine/Verbs/dotnet20sp2/script.js
Zemogiter: vcrun2003 also have those dll overrides on separate lines

ImperatorS79: Ok, but I find this ugly ^^.

Zemogiter: And why do I need to remove the dll overrides? It's not how winetricks does this.

Engines/Wine/Verbs/dotnet20sp2/script.js
Zemogiter: `dotnet20` have similiar structure (calling `removeMono` in every if/else)

Engines/Wine/Verbs/dotnet20sp2/script.js
ImperatorS79: Winetricks use an environment variable to only change the dlls while installing the verb.

Engines/Wine/Verbs/dotnet20sp2/script.js
ImperatorS79: Well, it is BAD ^^.

Zemogiter: How so?

Engines/Wine/Verbs/dotnet20sp2/script.js
Zemogiter: Will the latest commit suffice?

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -55,6 +55,9 @@ Wine.prototype.dotnet20sp2 = function () {
     }
 
     this.windowsVersion(osVersion);
+    this.overrideDLL()
+        .set("", dlls)
```
ImperatorS79: This is dangerous: you could remove other overrides !

Zemogiter: Unlikey, the `dlls` variable have only these overrides in it.

ImperatorS79: And what if other people have override dlls before installing the verb ?

ImperatorS79: Ok sorry, I have seen now what you have done. Clever way ^^ :+1:

Engines/Wine/Verbs/dotnet20sp2/script.js
Zemogiter: I tried to do it this way:
https://superuser.com/questions/978864/how-to-perform-unattended-installation-for-any-installer-using-wine

Zemogiter:
Tested again with the commits applied but the error remains.

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,79 @@
...
+        var setupFile32 = new Resource()
+            .wizard(this.wizard())
+            .url("https://download.microsoft.com/download/c/6/e/c6e88215-0178-4c6c-b5f3-158ff77b1f38/NetFx20SP2_x86.exe")
+            .checksum("22d776d4d204863105a5db99e8b8888be23c61a7")
+            .name("NetFx20SP2_x86.exe")
+            .get();
+
+
+        this.wizard().wait(tr("Please wait while {0} is installed...", ".NET Framework 2.0 SP2"));
+        this.run(setupFile32, ["--unattended", "/c:install.exe /q"], null, false, true);
```
ImperatorS79: That line should be  `this.run(setupFile, [setupFile, "/q", "/c:\"install.exe /q\""], null, false, true);`

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,79 @@
...
+        var setupFile64 = new Resource()
+            .wizard(this.wizard())
+            .url("https://download.microsoft.com/download/c/6/e/c6e88215-0178-4c6c-b5f3-158ff77b1f38/NetFx20SP2_x64.exe")
+            .checksum("a7cc6c6e5a4ad9cdf3df16a7d277eb09fec429b7")
+            .name("NetFx20SP2_x64.exe")
+            .get();
+
+
+        this.wizard().wait(tr("Please wait while {0} is installed...", ".NET Framework 2.0 SP2"));
+        this.run(setupFile64, ["--unattended", "/c:install.exe /q"], null, false, true)
```
ImperatorS79: Same as above.

Engines/Wine/Verbs/dotnet20sp2/script.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    this.windowsVersion(osVersion);
+    this.regedit().deleteValue("HKCU\\Software\\Wine\\DllOverrides", "ngen.exe");
```
ImperatorS79: I think you should add a *¨in front of each override if you want them to be deleted (phoenicis create them by default with * in front because it is a special settings in wine, but you should check to be sure)

Zemogiter:
All right the script is now working.

PR #951, opened 2019-05-20
Update Wine LATEST_STABLE_VERSION to 4.0.1

plata:


plata:
Note: do not merge before version is available.

qparis:
I think it is, isn't it?

ImperatorS79:
Yes, it is.

Issue #957, opened 2019-06-03
Add Minion Master


Kreyren:
Works on staging, requires UseTakeFocus disabled.

Graphical glitches present,

Steam required, investigate discord installer

Add DXV9

### Lutris installer
```yaml
files:
- d9vk: https://github.com/Joshua-Ashton/d9vk/releases/download/0.11/d9vk-0.11.tar.gz
game:
  appid: 489520
  arch: win64
  prefix: $GAMEDIR/prefix
installer:
- extract:
    dst: $CACHE/d9vk-0.11/
    file: d9vk
- copy:
    dst: $HOME/.local/share/lutris/runtime/dxvk/d9vk-0.11/
    src: $CACHE/d9vk-0.11/dxvk-release/
- task:
    arch: win64
    description: Creating Wine prefix.
    name: create_prefix
    prefix: $GAMEDIR/prefix
- task:
    app: usetakefocus=disabled
    arch: win64
    description: Disabling UseTakeFocus contributed by our God Kreyren (https://github.com/Winetricks/winetricks/pull/1246)
    name: winetricks
    prefix: $GAMEDIR/prefix
system:
  env:
    DXVK_HUD: 0
    DXVK_LOG_LEVEL: none
    DXVK_STATE_CACHE_PATH: $GAMEDIR/prefix
    __GL_SHADER_DISK_CACHE: 1
    __GL_SHADER_DISK_CACHE_PATH: $GAMEDIR/prefix
    __GL_THREADED_OPTIMIZATIONS: 1
wine:
  dxvk: true
  dxvk_version: d9vk-0.11
  version: ge-protonified-4.9-x86_64
```

ImperatorS79:
Why do you open an issue for that ? Just open a PR when you are ready or need for help ^^.

And what is the point of a "discord installer" ? Discord have a native linux version.

Kreyren:
> Why do you open an issue for that ? Just open a PR when you are ready or need for help ^^.

Because making installers for phoenicis is pita that never paid off in a way that end-user could play the game in my experience and so i'm prefering to share and track my test results here with the option of phoenicis contributors to add their results if wanted that can be then sent to WineHQ since WineHQ test results has to be approved and can't be editted after to be adapted once phoenicis can be used for compatibility layer management. Which based on current development should be any time now o.o else i'm getting triggered and deploying my own engine with runtime.

>  Why not rename this issue "Add Discord installer.. 

Creating new issue for that.

[...]

Issue #881, opened 2019-03-05
HIGH_PRIO: Unable to fetch *ANY* wine engine after staging x86 4.9 build


Kreyren:
Results in: 

![image](https://user-images.githubusercontent.com/11302521/53779792-defab400-3f01-11e9-8f76-d4ee80993585.png)

HOTFIX: Install wine manually.

Suspect bug?

_Originally posted by @Kreyren in https://github.com/PhoenicisOrg/scripts/pull/797#issuecomment-469525958_

plata:
Any error during Wine installation? What exactly did you do?

ImperatorS79:
@Kreyren Are there files in `.Phoenicis/engines/wine/staging-linux-x86/4.2/bin` ?

Zemogiter:
This error mentions wine staging 4.2 x86 but there is no such wine in Engines tab.
```
java.lang.RuntimeException: java.io.IOException: Cannot run program "/home/jonasz/.Phoenicis/engines//wine/staging-linux-x86/4.2/bin/wine" (in directory "/home/jonasz/.Phoenicis/containers/wineprefix/League of Legends"): error=2, Nie ma takiego pliku ani katalogu
	at jdk.scripting.nashorn/jdk.nashorn.internal.runtime.linker.JavaAdapterServices.wrapThrowable(JavaAdapterServices.java:285)
	at jdk.nashorn.javaadapters.org_phoenicis_engines_Engine.run(Unknown Source)
	at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$913$4947AAAZZA$\^eval\_.run(engines.wine.engine.object:178)
	at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$912$219A$\^eval\_$cu1$restOf.regedit#open(engines.wine.plugins.regedit:12)
	at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$899$287$\^eval\_$cu1$restOf.luna(engines.wine.verbs.luna:29)
	at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$858$493$\^eval\_$cu1$restOf.go(engines.wine.quick_script.installer_script:72)
	at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$826$603$\^eval\_.run(<eval>:18)
[...]

Issue #959, opened 2019-06-05
Cannot install any Steam app


Kreyren:
None of steam games are successfully installed on phoenicis

Actuall result is that phoenicis installs steam and halts on `Downloading runtime json`

```
[INFO ] org.phoenicis.tools.archive.Tar (l.172) - Creating output file /home/kreyren/.Phoenicis/engines/wine/upstream-linux-x86/4.0/./bin/winepath (493).
[INFO ] org.phoenicis.tools.archive.Tar (l.172) - Creating output file /home/kreyren/.Phoenicis/engines/wine/upstream-linux-x86/4.0/./bin/winedbg (493).
[INFO ] org.phoenicis.tools.archive.Tar (l.172) - Creating output file /home/kreyren/.Phoenicis/engines/wine/upstream-linux-x86/4.0/./bin/msiexec (493).
[INFO ] org.phoenicis.tools.archive.Tar (l.172) - Creating output file /home/kreyren/.Phoenicis/engines/wine/upstream-linux-x86/4.0/./bin/winefile (493).
[INFO ] org.phoenicis.tools.archive.Tar (l.172) - Creating output file /home/kreyren/.Phoenicis/engines/wine/upstream-linux-x86/4.0/./bin/winemaker (493).
Warning: Nashorn engine is planned to be removed from a future JDK release
[WARN ] org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO (l.56) - Installation ID (Mafia II_1559755098177) contains invalid characters, will remove them.
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
[WARN ] org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO (l.56) - Installation ID (Mafia II_1559755330515) contains invalid characters, will remove them.
Warning: Nashorn engine is planned to be removed from a future JDK release
[WARN ] org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO (l.56) - Installation ID (Mafia II_1559755338442) contains invalid characters, will remove them.
```

To reproduce install Mafia II on https://github.com/PhoenicisOrg/phoenicis/commit/15bc8a0f15afca5556cb332ee64b718ddd39155f

[...]

PR #932, opened 2019-05-09
Add D9VK verb

ImperatorS79:
D9VK is a direct3d9 implementation based on DXVK. I did not test he script, so if anyone can try it, it would be great ;-).

PR #960, opened 2019-06-05
Add MinionMasters

Kreyren:
DISCLAIMER: `main.png` is responsibility of phoenicis developers where it's beeing used under fair use.

Affected by https://github.com/PhoenicisOrg/scripts/issues/881#event-2391939521 -> Can not test.

Fixes: https://github.com/PhoenicisOrg/scripts/issues/957

Should work, bcs i don't do mistakes :p 

DO NOT MERGE! Not tested!


Kreyren:
`main.png` is missing and it won't be contributed by me cause of https://github.com/PhoenicisOrg/phoenicis/issues/1943

plata:
Both scripts are working? Does it make sense to provide a separate version for DXV9? I think that it might be confusing for the average user.

Kreyren:
> Both scripts are working? Does it make sense to provide a separate version for DXV9? I think that it might be confusing for the average user.

i'm not sure how DXV9 works so untill it's confirmed to be reliable i'm providing both options with preference on non-DXV9 once confirmed i will prefer DXV9 but still support other configurations.

plata:
@ImperatorS79 what do you think about this?

ImperatorS79:
I wonder if it should not be better to simply ask in the D9VK and DXVK verbs if the user wants to install it or not. The best would be to test vulkan support to do the decision (as DXVK brings **more** performance than wine3d)

There is a problem with jmimemagic currently preventing phoenicis to untar any .tar (see #932).

plata:
In any case, I think we should have an automatic way because many users will not know what DXVK is. Maybe the best solution for now is using only the script without DXV9. What do you think?

> There is a problem with jmimemagic currently preventing phoenicis to untar any .tar (see #932).

Does this mean that this script cannot work at all currently (including that it must be untested)?

ImperatorS79:
For any reasons, all script "untaring" things have started to not work :/.

plata:
Like stated before, we cannot have untested pull requests (or at least in draft if support is required). Therefore I close this until it is tested.

Kreyren:
@plata works on my end, but blocked https://github.com/PhoenicisOrg/scripts/issues/959

open the MR omg

plata:
Sorry but this is impossible. You're telling me that you cannot install any Steam app but have tested a Steam app at the same time?

Kreyren:
@plata *phoenicis can't install any steam apps..

 i installed it manually once phoenicis launched steam to test it. There are however some issue with steam that i'm investigating now and game has graphical glitches that are reproducable on random which i believe are caused either by my GPU configuration or DXVK.

Kreyren:
@plata https://i.imgur.com/S8S6EJY.jpg

PR #964, opened 2019-06-05
Add method to change container wine version in engine implementation 

ImperatorS79:
Apart from the fact that the engine are not sorted in the wizard and that it is a little bit difficult to find the version, it works !

EDIT: .sort() is good ^^.

https://github.com/PhoenicisOrg/phoenicis/pull/1970


ImperatorS79:
Ready for review.

plata:
I think the title should be adjusted. With this implementation, we can change the Wine version also from CLI etc. @madoar any idea for a better title?

ImperatorS79:
Add method to change container wine version in engine implementation ?

PR #990, opened 2019-06-06
Fix UplayScript documentation

Kreyren:
Fixes https://github.com/PhoenicisOrg/scripts/issues/989#issuecomment-499593708


plata:
Looks good but we must wait until the checks are finished.

Kreyren:
> Looks good but we must wait until the checks are finished.

ideally if you made the documentation less pita to contribute to i could update more things :p

plata:
I cannot see what was difficult/pita here.

PR #992, opened 2019-06-07
Update DXVK/Faudio/VK9: pass string instead of MenuItem

ImperatorS79:
It will never work as expected without that ^^.


plata:
Please always wait for a review and delete unused branches after merging.

PR #995, opened 2019-06-08
Added DXVK for Steam

Kreyren:
As refferenced on https://appdb.winehq.org/objectManager.php?sClass=version&iId=19444&iTestingId=103590:
image

Wine doesn't have sufficient DirectX 10 libs -> Using DXVK

### DIFF
--- a/Engines/Wine/QuickScript/Steam Script/script.js
+++ b/Engines/Wine/QuickScript/Steam Script/script.js
@@ -6,6 +6,7 @@ include("utils.functions.filesystem.extract");
 include("utils.functions.filesystem.files");
 include("engines.wine.verbs.luna");
 include("engines.wine.verbs.corefonts");
+include("engines.wine.verbs.dxvk");
 include("engines.wine.plugins.windows_version");
 
 
@@ -119,6 +120,7 @@ SteamScript.prototype.go = function () {
         .prefix(this._name, this._wineDistribution, this._wineArchitecture, this._wineVersion)
         .luna();
     wine.corefonts();
+    wine.DXVK(); // Required for big picture
 
     // Steam must be started once such that config.vdf is created (see fixCertificateIssue())
     setupWizard.wait(tr("Please follow the steps of the Steam setup. Then, wait until Steam is updated, log in and finally close Steam completely so the installation of \"{0}\" can continue.", this._name));
--- a/Applications/Games/Steam/Online/script.js
+++ b/Applications/Games/Steam/Online/script.js
@@ -1,4 +1,6 @@
 include("engines.wine.quick_script.online_installer_script");
+include("engines.wine.verbs.corefonts");
+include("engines.wine.plugins.windows_version");
 
 var installerImplementation = {
     run: function () {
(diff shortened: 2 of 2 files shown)


Engines/Wine/QuickScript/Steam Script/script.js
```diff
@@ -119,6 +120,7 @@ SteamScript.prototype.go = function () {
         .prefix(this._name, this._wineDistribution, this._wineArchitecture, this._wineVersion)
         .luna();
     wine.corefonts();
+    wine.DXVK(); // Required for big picture
```
ImperatorS79: While I could agree to add it for the steam app itself, it seems weird to add it here, because then all SteamScript will implicitly use DXVK, and I do not think it is a good idea.

Kreyren:
We might want to make a quick script for steam with bigpicture or some better way to handle this..

Engines/Wine/QuickScript/Steam Script/script.js
Kreyren: We can in theory load DXVK only for tenfoot (Big Picture) or depending on the implementation i was thinking about providing latest minus one version DXVK in phoenicis wine that can be overwritten by script on demand if game depends on different version.

plata: Also we need to keep in mind that not everybody will be able to run DXVK.

Kreyren: @plata elaborate

plata: To quote from the [DXVK README](https://github.com/doitsujin/dxvk):

> Notes on Vulkan drivers
Before reporting an issue, please check the Wiki page on the current driver status and make sure you run a recent enough driver version for your hardware.

> Online multi-player games
Manipulation of Direct3D libraries in multi-player games may be considered cheating and can get your account banned. This may also apply to single-player games with an embedded or dedicated multiplayer portion. Use at your own risk.

So even if we assume that everybody can use Vulkan (which is not the case), there are cases where DXVK cannot be used. I don't want anybody blaming us that his/her account in a multi-player game is banned. Therefore I think we cannot add DXVK by default anywhere (neither for Steam nor for Phoenicis Wine in general).

Kreyren: So we have to provide standalone DXVK libs for steam which doesn't affect any other apps.

ImperatorS79: Does it work if you copy dxvk libs next to steam.exe and set those to builtin for steam.exe only so that it does not affect other apps ?

plata: I have no clue. Maybe these questions should be raised to the DXVK devs. Or we simply continue to use DXVK on an per app basis only.
There's also one other point which might need consideration: It seems that Wine officially will go for the Wine D3D Vulkan implementation instead of DXVK. So maybe DXVK will become obsolete in the future.

ImperatorS79: Yes, but this is in the "far" future, and the two backend OpenGL/Vulkan will remains.

plata: That's true. However, isn't that a reason to use DXVK only per app? It will allow us to always implement what works best.

ImperatorS79: How do you do it is the question I was asking.

plata: That's my understanding as well.

Kreyren: afaik there is a way to use certain lib for wineapp only, investigating atm.. 

do not bother sujin with this since it's not DXVK question..

ImperatorS79:
Please reopen this of you found a way to only overrides DXVK for steam.

PR #1002, opened 2019-06-11
Update The Sims 3 (use Wine stable)

Zemogiter:
Previous used version is now crashing the Sims3LauncherW.exe making it impossible to play. Tried other versions and settled down on this one.

### DIFF
--- a/Applications/Games/The Sims 3/Local/script.js
+++ b/Applications/Games/The Sims 3/Local/script.js
@@ -13,8 +13,6 @@ var installerImplementation = {
             .author("Zemogiter")
             .category("Games")
             .executable("Sims3Launcher.exe", ["xgamma -gamma 1"])
-            .wineVersion("4.0.1")
-            .wineDistribution("upstream")
             .preInstall(function (wine){
                 wine.mfc42();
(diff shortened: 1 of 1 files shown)


Applications/Games/The Sims 3/Local/script.js
```diff
@@ -13,7 +13,7 @@ var installerImplementation = {
             .author("Zemogiter")
             .category("Games")
             .executable("Sims3Launcher.exe", ["xgamma -gamma 1"])
-            .wineVersion("3.19")
+            .wineVersion("4.0.1")
```
plata: Wouldn't it make sense to use the default (latest stable)?

Zemogiter: Maybe but what if they release a new stable 3 months from now and the script stops working?

plata: That can always happen. On the other hand, with this approach, somebody always must update the versions manually which is worse in my opinion.

Zemogiter: You have a point. I guess lack of action is worse than any action.

PR #1003, opened 2019-06-11
Fix runtime not being correctly added to the ldpath

ImperatorS79:
Btw @qparis the json from https://phoenicis.playonlinux.com/index.php/runtime?os=linux has not been updated.


qparis:
Why would it be updated?

ImperatorS79:
Because you should have updated the runtime ? ^^ https://github.com/PhoenicisOrg/phoenicis-winebuild/pull/91

PR #956, opened 2019-06-02
Read environment variables and trust level from shortcut

ImperatorS79:
This PR will also need a PR in phoenicis in the display/edit shortcut code.

However, I did not manage to send custom environment variable to the shortcut create method. The `environmentVar` variable is always undefined. I tried to send `[[DXVK_HUD, 1], [COUCOU, 27]]`.

It aims to fix #869 .

PR #1008, opened 2019-06-14
Refactor the ShortcutReader and WineShortcutReader Classes

madoar:
This PR refactors the `ShortcutReader` and `WineShortcutReader` classes to use the newer and better readable `class` syntax of JS. In the future I would like to change all our classes to use the new `class` syntax.


Engines/Wine/Shortcuts/Reader/script.js
```diff
@@ -1,122 +1,123 @@
...
 
-var _WineShortcutReader = function (shortcut) {
-    var that = this;
-    that._shortcutManager = Bean("shortcutManager");
-    that._libraryManager = Bean("libraryManager");
-    that._uiQuestionFactory = Bean("uiQuestionFactory");
-    that._winePrefixesDirectory = Bean("propertyReader").getProperty("application.user.containers") + "/" + WINE_PREFIX_DIR + "/";
+class WineShortcutReader {
+    constructor(shortcut) {
+        this.shortcut = shortcut;
```
madoar: We mainly access `this.shortcut` using `JSON.parse(this.shortcut.script)` maybe it makes sense to provide this value as either a getter method:

```javascript
get shortcutScript() {
   return JSON.parse(this.shortcut.script)
}
```

or as a public field.

What do you think @plata @ImperatorS79?

ImperatorS79: NO IDEA ^^.

ImperatorS79:
@madoar You did not look at my PR maybe that already try to do some changes to add multiple environment variables :/ #956.

Engines/Wine/Shortcuts/Reader/script.js
plata: I don't care too much about this detail but you're approach is fine for me.

madoar:
> - Why did you remove the `_` in front of members?

The `_` was used to mark "private" members, which don't exist in JS yet. I think that the `_` in front of variable-, function- and class-names looks strange. In addition I'm not sure whether we really have fields which absolutely mustn't be accessed from outside the class, therefore I removed them.

This change makes the code more in line with our Java code.

> - Does this syntax require any special JS environment which could possible introduce restrictions somewhere else?

Yes, it won't run on Nashorn. If we ignore Nashorn it should run on any actual JavaScript engine.

> - If we use this approach, we definitely need a documentation page describing it (to know how the classes should look like).

I don't see why? The `class` keyword is a normal JavaScript keyword, which has been introduced in ECMAScript 2015. A documentation to this keyword can be found on any [JS language documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes).

plata:
> The _ was used to mark "private" members, which don't exist in JS yet. I think that the _ in front of variable-, function- and class-names looks strange. In addition I'm not sure whether we really have fields which absolutely mustn't be accessed from outside the class, therefore I removed them.

At least it was an indication that the member should not be used from outside. @qparis what do you think about this.

> I don't see why? The class keyword is a normal JavaScript keyword, which has been introduced in ECMAScript 2015.

We want a common style in our repository. Therefore it makes sense to document which style we want to use.

qparis:
It's more a matter of convention. Personally, I like to encourage people writing as much methods as possible (instead of comments). This convention helps to keep a clear class contract

plata:
For me either one is fine. We should just take a decision and then write that down in the documentation. So I'll leave that to you and @madoar.

madoar:
Ok, I'll add a short paragraph about the used `class` syntax to our documentation.
Still we'll need to find a decision for the "private" members. I'm not a fan of adding an `_` to the method/field/class names. 

Long-term I think the "private" problem will be solved by a newer JavaScript version, see https://github.com/tc39/proposal-private-methods and https://github.com/tc39/proposal-class-fields

madoar:
@plata see be16754019d26b6edd857b9d74dc9d309be8b8f3 for the documentation page

PR #1010, opened 2019-06-15
Refactor the Quick Script Classes

madoar:
This PR refactors the quick script classes with the newer and better understandable JavaScript class syntax. In addition this PR replaces the `var` keyword with the `const` or `let` keyword depending on whether the variable is a constant or not.

@plata @qparis @ImperatorS79 I've tested the changes in this PR by reinstalling Notepad++. Because this only uses three of the quick script classes it would be nice if you can test some of the others, which I'm unable to (origin, gog, local, custom, uplay, zip)

### DIFF
--- a/Engines/Wine/QuickScript/Custom Installer Script/script.js
+++ b/Engines/Wine/QuickScript/Custom Installer Script/script.js
@@ -1,14 +1,12 @@
 include("engines.wine.quick_script.installer_script");
 
-function CustomInstallerScript() {
-    InstallerScript.call(this);
+class CustomInstallerScript extends InstallerScript {
+    constructor() {
+        super();
+    }
+
+    installationCommand(installationCommand) {
+        this._installationCommand = installationCommand;
+        return this;
+    }
 }
-
-CustomInstallerScript.prototype = Object.create(InstallerScript.prototype);
-
-CustomInstallerScript.prototype.constructor = CustomInstallerScript;
-
-CustomInstallerScript.prototype.installationCommand = function (installationCommand) {
-    this._installationCommand = installationCommand;
-    return this;
-};
--- a/Engines/Wine/QuickScript/GoG Script/script.js
+++ b/Engines/Wine/QuickScript/GoG Script/script.js
@@ -3,108 +3,107 @@ include("utils.functions.net.download");
 include("engines.wine.engine.object");
 include("engines.wine.verbs.gdiplus");
 
-function GogScript() {
-    QuickScript.call(this);
-}
+class GogScript extends QuickScript {
+    constructor() {
+        super();
+    }
 
-GogScript.prototype = Object.create(QuickScript.prototype);
+    /**
+     * Sets one setup file name so that the script can fetch it from gog.com
+     * @param {string} setupFileName The setup file name
+     * @returns {GogScript} This
+     */
+    gogSetupFileName(setupFileName) {
+        this._setupFileNames = [setupFileName];
+        return this;
+    }
 
-GogScript.prototype.constructor = GogScript;
+    /**
+     * Sets the setup file(s) name so that the script can fetch it from gog.com
+     * @param {string[]} setupFileNames The setup file name(s)
+     * @returns {GogScript} This
+     */
+    gogSetupFileNames(setupFileNames) {
+        this._setupFileNames = setupFileNames;
+        return this;
+    }
 
-/**
- * Sets one setup file name so that the script can fetch it from gog.com
- * @param {string} setupFileName The setup file name
- * @returns {GogScript} This
- */
-GogScript.prototype.gogSetupFileName = function (setupFileName) {
-    this._setupFileNames = [setupFileName];
-    return this;
-}
+    /**
+     * Presents a Gog.com login window to the user, login to its account and return a token that can be used later.
+     * Stores the tocken in a parameter
+     * @param {SetupWizard} setupWizard The setupWizard to use
+     * @returns {GogScript} This
+     */
+    loginToGog(setupWizard) {
+        const browserWindow = setupWizard.createBrowser(tr("Please login to your GoG.com account so that we can download the game for you:"));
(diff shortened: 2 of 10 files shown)


ImperatorS79:
I cannot test anything since I cannot launch any script since the graalVM switch. Does it work with nashorn ?

Engines/Wine/QuickScript/Custom Installer Script/script.js
```diff
@@ -1,14 +1,12 @@
 include("engines.wine.quick_script.installer_script");
 
-function CustomInstallerScript() {
-    InstallerScript.call(this);
+class CustomInstallerScript extends InstallerScript {
+    constructor() {
+        super();
+    }
+
+    installationCommand(installationCommand) {
```
madoar: I don't like the names of our `QuickScript` methods. They are named like accessor methods and fields but function as fluid API methods, which return the object itself. To make this more clear and to clear naming conflicts between the method name and the underlying field I propose to rename the methods to:

```javascript
withInstallationCommand(installationCommand) {
   this.installationCommand = installationCommand;
   return this;
}
```

Again this would also be more in line with our Java implementation.

Engines/Wine/QuickScript/GoG Script/script.js
```diff
@@ -3,108 +3,107 @@ include("utils.functions.net.download");
...
+
+        const currentUrl = browserWindow.getCurrentUrl();
+        const code = currentUrl.split("code=")[1].split("&")[0];
+
+        const tokenUrl = "https://auth.gog.com/token?client_id=46899977096215655&client_secret=9d85c43b1482497dbbce61f6e4aa173a433796eeae2ca8c5f6129f2dc4de46d9&grant_type=authorization_code&code=" + code + "&redirect_uri=https%3A%2F%2Fembed.gog.com%2Fon_login_success%3Forigin%3Dclient";
+        this._token = new Downloader().url(tokenUrl).json();
+        return this;
+    }
+
+    _downloadSetupFile(setupWizard, setupFileName, tmpDirectory) {
```
madoar: I don't think we need to name our "private" methods with a prefix `_`, because it can't block someone from using it if he really wants to. If we want to make it clear that the method is intended as a private method I think we should add it as a comment before the method, something like:

```javascript
/**
 * *** INTERNAL API ***
 * ...
 */
```

Engines/Wine/QuickScript/Quick Script/script.js
```diff
@@ -1,132 +1,136 @@
...
+        this._wineDistribution = wineDistribution;
+        return this;
+    }
+
+    wineVersion(wineVersion) {
+        this._wineVersion = wineVersion;
+        return this;
+    }
+
+    wineUserSettings(wineUserSettings) {
```
madoar: If we use a different name for the fluid API methods, we don't need separate code for the value access anymore:

```javascript
withWineUserSettings(wineUserSettings) {
   this.wineUserSettings = wineUserSettings;
   return this;
}

...
// accessing wineUserSettings
const userSettings = quickScript.wineUserSettings;
```

Engines/Wine/QuickScript/Quick Script/script.js
```diff
@@ -1,132 +1,136 @@
...
+    preInstall(preInstall) {
+        this._preInstall = preInstall;
+        return this;
+    }
+
+    /**
+     * creates shortcut
+     * @param {string} [prefix] prefix name
+     */
+    _createShortcut(prefix) {
```
madoar: Why is this method named with an `_`? It is always used from the outside I believe.

madoar:
> I cannot test anything since I cannot launch any script since the graalVM switch. 

This is interesting. How many repositories have you added to Phoenicis? Are they all based on the current master branch of the Scripts repository?

> Does it work with nashorn ?

No it doesn't work with Nashorn. Nashorn uses a really old standard of JavaScript.

ImperatorS79:
I will have to retest with master branch to see.

plata:
Let's discuss the style in the other issue first and document it. Then we can apply it to everything else.

ImperatorS79:
Side note: my issue comes from missing libs in the runtime, I am currently investigating it ^^.

Engines/Wine/QuickScript/Custom Installer Script/script.js
qparis: Agreed

Engines/Wine/QuickScript/GoG Script/script.js
qparis: I think we do have a private keyword in JS class. I suggest we use it.

Engines/Wine/QuickScript/Quick Script/script.js
qparis: Hum it is not. If I remember well, it is only called internally by run() or go()

Engines/Wine/QuickScript/GoG Script/script.js
madoar: Are you sure JS contains a private keyword? I'm not aware of such a keyword.

madoar:
@plata I think we can merge this PR, what do you think?

Engines/Wine/QuickScript/Custom Installer Script/script.js
plata: Can you open a new issue to track this?

Engines/Wine/QuickScript/GoG Script/script.js
plata: Can you open a new issue to track this?

Engines/Wine/QuickScript/Custom Installer Script/script.js
madoar: See @1021

Engines/Wine/QuickScript/GoG Script/script.js
madoar: See #1022

PR #1011, opened 2019-06-15
Simplify the Installer Scripts

madoar:
This PR aims to simplify the Installer Scripts by removing the wrapper `run` method.
Currently this PR changes only the Notepad++ script for discussion purposes.

This PR requires https://github.com/PhoenicisOrg/phoenicis/pull/2002 to work


qparis:
Very elegant

ImperatorS79:
This was how script used to look like some times ago. I did not remember why it changes. Any reason why the .go(); is not here as well in the design ?

plata:
If I understand correctly, this works because of duck typing, right? Can we verify that this approach will also work for verbs, plugins, engines etc. before we decide to merge this? If it does, I like it a lot because the scripts are easier to use/implement.

madoar:
> If I understand correctly, this works because of duck typing, right?

Yes, in the current simple form Graal.js applied the JS object to the interface by duck typing.

> Can we verify that this approach will also work for verbs, plugins, engines etc. before we decide to merge this?

The **engines**, **settings** and **tools** themselves can be handled similar I think. They already provide methods mirroring the Java interface. This only requires the extra-step to create an object from the class in Java, which is not necessary for the installers. The engine extensions (**plugins**, **verbs**) will be harder, because they don't provide their own uniform interface, instead they are extending the interface of the engine.

madoar:
> This was how script used to look like some times ago. I did not remember why it changes.
We changed this previously because we didn't want to execute the installer automatically when "loading" the script. Therefore we added a short wrapper, which needs to be explicitly executed to run the script.

> Any reason why the .go(); is not here as well in the design ?

I discovered that what we wanted to initially achieve, i.e. don't execute the installer directly, can also be achieved by removing the `.go()` method call. The `.go()` method call is responsible for executing the installer. So if I remove that call and execute it somewhere on the Java side I achieve the same effect.

plata:
> The engines, settings and tools themselves can be handled similar I think.

Can you verify this with a proof of concept?

> The engine extensions (plugins, verbs) will be harder

Can you check if we can use the same approach nevertheless? I think it would not be a good idea to follow two different approaches.

madoar:
> Can you check if we can use the same approach nevertheless? I think it would not be a good idea to follow two different approaches.

What do you mean by the same approach?

Engines/Wine/Verbs/adobeair/script.js
```diff
@@ -28,16 +28,20 @@ Wine.prototype.adobeair = function () {
 /**
  * Verb to install adobeair
  */
-var verbImplementation = {
-    install: function (container) {
-        var wine = new Wine();
+class AdobeAirVerb {
```
madoar: @plata @qparis this approach should work for all types of scripts that return a class which needs to be instantiated on the Java side.

plata: Can you try also with `Engine`. I tried myself but it didn't work. Either I did something wrong or I have a different issue here.

plata:
> What do you mean by the same approach?

That all our implementations of Java interfaces in Javascript should follow the same style.

Engines/Wine/Verbs/adobeair/script.js
madoar: Engine should work the same but will need more effort:

Ideally I would like to merge both the engine object and the engine implementation into a single JavaScript class. Afterwards we should be able to work with it using the same approach.

I don't want to do these changes in this PR, because I can't say yet what else in the scripts needs to be changed to make them work with a unified object.

madoar: How did you try to make it work?

plata: Basically I tried to replace the `Java.extend` syntax by `class`. It should not be part of this pr (also the verbs). But in my opinion, we should really be sure that it will work. So what I would like is: one pr for each type (installer scripts, verbs, plugins, engines ...) with one adjusted script. Review that and then apply the changes to all scripts of that type. Then merge.

madoar: I agree for the multi PR approach. On the Java side I think we will need a shared change (i.e. the semantic change of the `include` command). I would like to provide this as a separate PR which we can merge as is, because it shouldn't affect the status quo.

About your experiments:
What error did you get?

plata: Stuff like: `PolyglotMap cannot be used as array` (in Wine#run when the `ProcessBuilder` is called). No clue if it's related at all but before my changes this was working.

plata:
@madoar if you have tested the functionality with new approach, I think the PRs can be extended for the remaining classes and then merged.

madoar:
I've formatted the installers with:

```
npm install -g prettier
prettier --print-width 120 --use-tabs --write **/script.js
```

Suggestions for better JS formatting are always open

Applications/Internet/Internet Explorer 6.0/Online/script.js
```diff
@@ -10,91 +10,159 @@ include("utils.functions.apps.resources");
 include("engines.wine.verbs.msls31");
 
 var installerImplementation = {
```
madoar: I couldn't change this script to the new style because it doesn't use a quick script. @plata @ImperatorS79 do you have an idea how we could transform this to an JS object?

Applications/Internet/Internet Explorer 7.0/Online/script.js
```diff
@@ -10,232 +10,311 @@ include("engines.wine.verbs.sandbox");
 include("engines.wine.plugins.windows_version");
 
 var installerImplementation = {
```
madoar: I couldn't change this script to the new style because it doesn't use a quick script. @plata @ImperatorS79 do you have an idea how we could transform this to an JS object?

madoar:
In addition I couldn't change the local `Space Colony` script, because it makes use of `.go(["/q"])`, which I don't call from Java. Any ideas?

ImperatorS79:
What is this call doing ? Launching the installer with an argument ?

madoar:
Fun fact:
The method `go` does not accept any parameters, see:

https://github.com/PhoenicisOrg/scripts/blob/968b8ee4fd1801d32ed444d7439c09689356b0c3/Engines/Wine/QuickScript/Installer%20Script/script.js#L12

ImperatorS79:
AH!

Applications/Internet/Internet Explorer 6.0/Online/script.js
plata: Can't you write it similarly to https://github.com/PhoenicisOrg/scripts/blob/master/Engines/Wine/QuickScript/Installer%20Script/script.js (except that it does not `extend QuickScript`)?

plata:
Do the Space Colony script without parameters for `go`. Don't think it will make any difference.

Applications/Internet/Internet Explorer 6.0/Online/script.js
madoar: Ok I've now introduced a `DummyInstallerScript` class, which is used by the IE installers.

As a note:
Both IE scripts are broken, independent of the changes done in this PR. IE 6 fails when downloading some file. IE 7 fails because of a syntax error in the script. 

The IE 7 syntax error I fixed, but after completing the installation I was unable to start it.

madoar:
Ok, I think I converted all installer scripts.

Applications/Internet/Internet Explorer 6.0/Online/script.js
plata: What's the advantage of the `DummyInstallerScript` compared to directly implementing the `go` method? Originally, I was thinking of something like:
```Javascript
class Ie6Installer {
    go() {
        ...
    }
}
new Ie6Installer();
```

plata:
Check Codacy. You must run ESLint to fix the formatting.

Applications/Internet/Internet Explorer 6.0/Online/script.js
madoar: There is not really an advantage. It just makes it simpler to read the file. In addition it is more in line with the style used for the other installers, which only contain a `new XYZInstallerScript()...` statement.

plata: Ok. I'm fine with using the approach but I really don't like the name nor the place. `DummyInstaller` somehow suggests that it's not a "real" installer. Also, it can be used for all engines (not just Wine) so it should not be in `Wine/QuickScripts`.

madoar: I'm not happy with introducing the class at all. In theory the IE installers should be some kind of `OnlineInstallerScript`s because they download an installer. In addition the installers don't work (independent of this PR). The best solution in my eyes would be to rewrite the IE installers to make use of `OnlineInstallerScript`.

plata: They are a showcase of how you can write completely free scripts (we explicitly want to support that).

madoar: I understand, then I think we should go with my `DummyInstallerScript` approach and give it a better name and move it a folder up to make it independent of Wine. 

Still we have the issue that the IE installers don't seem work.

plata: How about `Utils/Functions/Apps/Installers` and `PlainInstaller`? I currently do not care if the script works (if it was broken before already, this is not related to this pr).

madoar: Ok I have renamed `DummyInstallerScript` to `PlainInstaller`

plata:
@qparis if you are ok with the `PlainInstaller`, this can be merged.

madoar:
@plata @qparis summarized we can merge all the refactor engines/tools/settings/installers PRs right? (both in the phoenicis and the scripts repository)

plata:
Yes.

qparis:
Yes

PR #1015, opened 2019-06-16
Simplify the Verb Scripts

madoar:
This PR aims to simplify the Verb Scripts by changing the Verb wrapper to a normal JavaScript class
Currently this PR changes only the Adobe Air verb script for discussion purposes.

Requires https://github.com/PhoenicisOrg/phoenicis/pull/2006


plata:
What I still find a bit ugly is that you cannot switch the order of `Wine.prototype.adobeair` and `AdobeAirVerb`. I do not have any idea how this could be avoided though. @qparis do you?

madoar:
If we know the name of the verb implementation class beforehand (in Java) then we can reference it directly without requiring its definition as the last instruction. In theory we could give every verb class the same name, but I don't want to do this. Mainly because this makes it harder to differentiate the verbs by only looking at them and because it seems wrong to me.

madoar:
Found a missing verb implementation for `dotnet20`. Is this intentional?

Engines/Wine/Verbs/vcrun2003/script.js
```diff
@@ -22,29 +23,29 @@ Wine.prototype.vcrun2003 = function () {
         "msvcp71",
         "mfc71"
     ];
+
     var that = this;
```
madoar: It seems like the verb scripts contain this assignment alot. 
@plata @ImperatorS79 do you know if this is still necessary with GraalVM?

madoar:
I should have transformed all verbs now.

Engines/Wine/Verbs/vcrun2003/script.js
ImperatorS79: Normally yes, because without that you cannot access `wine` methods inside the `foreach`, however this is worth testing (just install the verb in dummy container to see).

Engines/Wine/Verbs/Remove Mono/script.js
```diff
@@ -3,12 +3,12 @@ include("utils.functions.filesystem.files");
...
-* @returns {Wine} Wine object
-*/
+ * Verb to remove mono
+ * 
+ * @returns {Wine} Wine object
+ */
 Wine.prototype.removeMono = function () {
-    if (this.uninstall("Mono"))
-    {
+    if (this.uninstall("Mono")) {
```
plata: Why doesn't ESLint handle this?

madoar: I don't know. I performed the formatting in this PR with Visual Studio Code.

Engines/Wine/Verbs/vcrun2003/script.js
madoar: Ah I remember reading something about differences with the scope when using `function(args) {...}` compared to `(args) => {...}`. 

If I remember correctly the first variant automatically sets the scope inside the function to `this` (e.g. prototype constructors), while the second one does not.

plata:
Codacy is broken.

PR #1016, opened 2019-06-16
Simplify the Settings Scripts

madoar:
This PR aims to simplify the Settings Scripts by changing the Settings wrapper to a normal JavaScript class
Currently this PR changes only the always offscreen settings script for discussion purposes.


Engines/Wine/Settings/always offscreen/script.js
```diff
@@ -4,45 +4,50 @@ include("engines.wine.plugins.regedit");
...
-    _registryValues: ["", "disabled", "enabled"],
-    getText: function () {
+class AlwaysOffscreenSetting {
+    constructor() {
+        this.options = [tr("Default"), tr("Disabled"), tr("Enabled")];
+        // values which are written into the registry, do not translate!
+        this.registryValues = ["", "disabled", "enabled"];
+    }
+
+    getText() {
```
madoar: I think it would be a good idea to change this method to a JavaScript getter method:

```javascript
get text() {
   return tr("Always offscreen");
}
```

What do you think?

Engines/Wine/Settings/always offscreen/script.js
```diff
@@ -4,45 +4,50 @@ include("engines.wine.plugins.regedit");
...
         return tr("Always offscreen");
-    },
-    getOptions: function () {
-        return this._options;
-    },
-    getCurrentOption: function (container) {
-        var currentValue = new Wine()
+    }
+
+    getOptions() {
```
madoar: I'm not sure whether this method is actually necessary. If we're lucky Graal.js automatically maps the getter method to the `options` field accessor.

Engines/Wine/Settings/always offscreen/script.js
```diff
@@ -4,45 +4,50 @@ include("engines.wine.plugins.regedit");
...
+        return this.options;
+    }
+
+    getCurrentOption(container) {
```
madoar: @plata do you think it makes sense to move the `container` argument to the constructor?
This would allow us to change this method to a normal JavaScript getter method:

```javascript
get currentOption() {
   ...
}
```

Engines/Wine/Settings/always offscreen/script.js
```diff
@@ -4,45 +4,50 @@ include("engines.wine.plugins.regedit");
...
-        var index = Math.max(this._registryValues.indexOf(currentValue), 0);
-        return this._options[index];
-    },
-    setOption: function (container, optionIndex) {
+        const index = Math.max(this.registryValues.indexOf(currentValue), 0);
+
+        return this.options[index];
+    }
+
+    setOption(container, optionIndex) {
```
madoar: Similar to `getCurrentOption` this method could be changed to a normal JavaScript setter method if we move the container argument to the constructor:

```javascript
set currentOption(optionIndex) {
   ...
}
```

Engines/Wine/Settings/always offscreen/script.js
plata: I'm not that familiar with Javascript getters. Any advantages (except that it looks nicer). Can it still be called the same way from Java?

Engines/Wine/Settings/always offscreen/script.js
plata: Can you try?

Engines/Wine/Settings/always offscreen/script.js
plata: I'm currently not quite sure where the objects are created. If they are created every time you select a container, we can do this. If they are created once and only called for different containers, it will not work.

Engines/Wine/Settings/always offscreen/script.js
plata: Same like above: what would be the benefits?

Engines/Wine/Settings/always offscreen/script.js
madoar: > I'm not that familiar with Javascript getters. Any advantages (except that it looks nicer).

No, this is only a cosmetical change.

> Can it still be called the same way from Java?

I still need to check. I just want to discuss this as early as possible

plata: I suggest to implement the simplification first without this and then have a look at it again.

madoar: I've just tried if `get text()` can be used as a synonym for `getText` in the Java interface, but it doesn't seem to work. I've opened an [issue](https://github.com/graalvm/graaljs/issues/181) about this in the GraalJS GitHub repository to see whether this is intentional or not.

PR #1017, opened 2019-06-16
Simplify the Tool Scripts

madoar:
This PR aims to simplify the Tool Scripts by changing the Tool wrapper to a normal JavaScript class
Currently this PR changes only the configure wine tool script for discussion purposes.


Engines/Wine/Tools/Configure Wine/script.js
```diff
@@ -2,14 +2,15 @@ include("engines.wine.engine.object");
...
-*/
-var toolImplementation = {
-    run: function (container) {
+ */
+class ConfigureWineTool {
+    constructor() {
+        // do nothing
+    }
+
+    run(container) {
```
madoar: I believe `execute` to be a better name instead of `run`. @plata @qparis @ImperatorS79 what do you think?

plata: No preference from my side.

qparis: Run is fine IMHO

madoar:
I should have transformed all tools now.

Engines/Wine/Tools/Wine Task Manager/script.js
```diff
@@ -1,15 +1,16 @@
...
+    run(container) {
         new Wine()
             .prefix(container)
             .run("taskmgr", [], null, false, true);
     }
-};
-
-/* exported Tool */
-var Tool = Java.extend(org.phoenicis.engines.EngineTool, toolImplementation);
+}
```
plata: new line at EOF

madoar: Done

PR #1018, opened 2019-06-20
Simplify the Engine Scripts

madoar:
This PR aims to simplify the Engine Scripts by changing the Engine wrapper to a normal JavaScript class.


Engines/Wine/Engine/Implementation/script.js
```diff
@@ -69,8 +72,8 @@ var engineImplementation = {
                 wizard.close();
             }
         }
-    },
-    _installWinePackage: function (setupWizard, winePackage, localDirectory) {
+    }
+    _installWinePackage(setupWizard, winePackage, localDirectory) {
```
plata: Do you want to keep the `_` notation here?

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -182,7 +185,7 @@ var engineImplementation = {
                             .url(archive.url)
                             .message(tr("Downloading x86 runtime..."))
                             .checksum(archive.sha1sum)
-                            .to(that._wineEnginesDirectory + "/TMP/" + archive.url.substring(archive.url.lastIndexOf('/')+1))
+                            .to(that._wineEnginesDirectory + "/TMP/" + archive.url.substring(archive.url.lastIndexOf('/') + 1))
```
plata: I wonder why ESLint did not complain about all this before.

Engines/Wine/Engine/Implementation/script.js
madoar: No. I didn't change the notation yet because the class is quite long and I only wanted to get it working for now. Long-term I think we should change this

plata:
I think your last commit is the place where I got issues when I tried this myself.

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -347,13 +347,9 @@ class WineEngine {
...
         var wineBinary = this.getLocalDirectory(subCategory, version) + "/bin/wine";
-        var StringArray = Java.type('java.lang.String[]');
-        var command = new StringArray(2 + args.length);
-        command[0] = wineBinary;
-        command[1] = executable;
-        java.lang.System.arraycopy(args, 0, command, 2, args.length);
+        var command = [wineBinary, executable].concat(args);
         var ProcessBuilderClass = Java.type('java.lang.ProcessBuilder');
-        var processBuilder = new ProcessBuilderClass(command);
+        var processBuilder = new ProcessBuilderClass(Java.to(command, Java.type("java.lang.String[]")));
```
plata: Is there a rule for `"java.lang.String[]"` vs `'java.lang.String[]'`. I think we should be consistent, see e.g. `'java.lang.ProcessBuilder'`.

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -335,13 +347,9 @@ var engineImplementation = {
         this.install(subCategory, version);
 
         var wineBinary = this.getLocalDirectory(subCategory, version) + "/bin/wine";
-        var StringArray = Java.type('java.lang.String[]');
-        var command = new StringArray(2 + args.length);
-        command[0] = wineBinary;
-        command[1] = executable;
-        java.lang.System.arraycopy(args, 0, command, 2, args.length);
```
madoar: @plata I think the problem you encountered previously was because of this line. 
The issue was that `args` is a JavaScript list, while `command` is a Java array. `System#arraycopy` doesn't support that signature, therefore the error.

I fixed that problem, but I think we should remove all direct calls to Java APIs like `ProcessBuilder`, to be sure to prevent the similar issues from occurring.

plata:
To fix Codacy: https://github.com/PhoenicisOrg/scripts/blob/662e0d434131403968dc407b38118e4e79ec922c/.eslintrc.yml#L62

Engines/Wine/Engine/Implementation/script.js
plata: Agree. Also @qparis mentioned this before. Can you open an issue to track that?

Engines/Wine/Engine/Implementation/script.js
madoar: What do you mean by "java.lang.String[]" vs 'java.lang.String[]'?

plata: single vs double quotes

madoar: No there is not, I simply copied the `Java.type(...)` command from somewhere else^^

madoar:
`Engine: false` or `WineEngine: false`?

Engines/Wine/Engine/Implementation/script.js
plata: Nevertheless I think we should always use the same. Lets say single quotes, ok?

plata:
`WineEngine`.

PR #1034, opened 2019-06-23
Refactor "filesystem.files" utility functions

madoar:
This PR refactors the `utils.functions.filesystem.files` utility functions.

Requires https://github.com/PhoenicisOrg/phoenicis/pull/2019


Utils/Functions/Filesystem/Files/script.js
```diff
@@ -2,202 +2,205 @@ var fileAnalyser = Bean("fileAnalyser");
...
+		this.checksumCalculator = Bean("checksumCalculator");
+		this._method = "SHA";
+	}
+
+	/**
+	 * sets wizard
+	 * @param {SetupWizard} wizard setup wizard
+	 * @returns {Checksum} Checksum object
+	 */
+	wizard(wizard) {
```
madoar: I think we should rename the fluid style API method to the `withX` syntax

qparis: Concerning wizard, I'm not sure. Even though, sometimes, it is useful to have the same name for getting and setting the object

madoar: As far as I understand the set parameters are all internal/private parameters right?
If this is the case we don't need a getter method right?

qparis: I think that there are some case when you may want to pass the wizard from an object to another

madoar: The issue is, that the methods not real setter methods, because they return `this`.

In my opinion it would be better to use:

```javascript
withWizard(wizard) {
   this.wizard = wizard;
}
```

Afterwards it is automatically possible to access the given wizard via:

```javascript
checksum.wizard
// or from inside the Checksum class
this.wizard
```

I don't think that this will lead to problems.

qparis: This is not a real issue IMHO, it is a good pattern to allow chaining stuff. 

I don't really like accessing property from the outside

madoar: Even in JavaScript?

qparis: why not?

madoar: I understand I thought you meant accessor methods, but in this case we're talking directly about accessing the property. Normally I would agree that allowing direct access to the property is bad, but I'm not really a friend of providing an imaginary name for a "private" property and then allowing access to it via a getter accessor method. If you think this is a cleaner solution we can also go with that.

I'm not suggesting to perform these changes in this PR, because it is simply out of scope

plata:
Please fix Codacy before merging.

madoar:
@plata I am not sure how to fix these errors? The functions are used in other places and `'Checksum' is defined but never used. (no-unused-vars)` leads to its own problems because this will apply to all refactored verbs, engines, tools and settings because they define a class which is only used from Java. 

I guess the way to "fix" these errors is to add them to the exclusions in `.eslintrc.yml` but I don't think this is a good idea long-term, because every new verb, engine, tool and script requires an additional exclusion.

plata:
There used to be `// eslint-disable-line no-unused-vars` comments in the file but you delete those which is why Codacy is failing now.

PR #1035, opened 2019-06-23
Fix some Issues with ShortcutReader

madoar:
This PR fixes some issues with the `ShortcutReader` that occur when you try to delete a shortcut.

Don't ask me why the formatting has changed. 
`eslint -c .eslintrc.yml --fix ./**/*.js` doesn't "fix" it.


Engines/Wine/Shortcuts/Reader/script.js
```diff
@@ -76,10 +73,10 @@ class ShortcutReader {
     }
 
     /**
-     * sets shortcut
-     * @param {string} shortcut shortcut
-     * @returns {void}
-     */
+	 * sets shortcut
```
plata: The indentation of these comments looks weird.

madoar: It seems like I had some custom settings on my PC that were the reason for this. Should be fixed now.

PR #1040, opened 2019-06-28
Update D9VK (fix OS detection, added new versions)

Zemogiter:
### Description
Per @plata request I checked if other scripts that need to detect host OS at some point have the same problem as DXVK.
### What works
Everything
### What was not tested
Nothing
### Test
- Operating system (and linux kernel version): Ubuntu 19.04 x64
- Hardware (GPU/CPU):
i7-7700k, GTX 1080 ti
### Ready for review


Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -11,10 +11,14 @@ include("utils.functions.filesystem.files");
...
 Wine.prototype.D9VK = function (d9vkVersion) {
     var operatingSystemFetcher = Bean("operatingSystemFetcher");
+    var uiQuestionFactory = Bean("uiQuestionFactory");
     print("NOTE: Wine version should be greater or equal to 3.10");
-    if (operatingSystemFetcher.fetchCurrentOperationSystem() != "Linux")
+    if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
     {
-        this.wizard().message(tr("D9VK might not work correctly on macOS. This is depending on Metal api support and MoltenVK compatibility layer advancement"));
+        uiQuestionFactory.create(
+				tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Do you want to continue? Chosing yes will skip DXVK verb  installation and continue with other verbs. Chosing no will quit script installation.", winePrefix),
```
plata: chosing → choosing, remove duplicate space in "verb   installation", what is `winePrefix` supposed to do?

Zemogiter: Not sure why it's there, that was in the shortcut reader script.

plata: The shortcut reader uses "{0}" in the string. The `tr` will fill this with the variable.

Zemogiter: So should I remove `winePrefix`?

plata: Depends what you want to achieve. You can also insert "${0}" in the string.

plata: Not like this. Let me explain
```Javascript
const myVariable = 5;
print(tr("This is my variable: ${0}", myVariable));
```
will print
```
This is my variable: 5
```

Zemogiter: Ok I guess we dont need that.

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -12,17 +12,22 @@ include("utils.functions.filesystem.files");
...
+    var uiQuestionFactory = Bean("uiQuestionFactory");
     print("NOTE: Wine version should be greater or equal to 3.10");
-
-    if (operatingSystemFetcher.fetchCurrentOperationSystem() != "Linux") {
-        this.wizard().message(tr("D9VK might not work correctly on macOS. This is depending on Metal api support and MoltenVK compatibility layer advancement"));
+    if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
+    {
+        uiQuestionFactory.create(
+				tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Do you want to continue? Choosing yes will skip DXVK verb installation and continue with other verbs. Choosing no will quit script installation."),
+				() => return this;
```
madoar: This looks like a syntax error to me:
1) when using the arrow syntax without `{...}` I don't think you are allowed to use `;` at the end
2) I don't think that this `return` allows you to exit the outer method (i.e. `Wine.D9VK(d9vkVersion)`), because it is part of an inner function

Have you tested this call to be sure it that it works as it should?

Zemogiter: No because I don't have a mac so the statement in that if is always false.

plata:
Assigned to @qparis to check on Mac.

plata:
@qparis could you check so that we can get this out of the way?

Engines/Wine/Verbs/D9VK/script.js
madoar: The example assigns the lambda function (i.e. `([a, b] = [1, 2], {x: c} = {x: a + b}) => a + b + c;`) to a variable (i.e. `var f`). This assignment is terminated with the `;` and not the lambda expression itself.

Have you tried executing `eslint -c .eslintrc.yml --fix ./**/*.js` to see if it shows any issues with the script?

Zemogiter: I get this:
```
  16:11  error  Parsing error: Unexpected token return
```

madoar: This means that there is a syntax error, because `return` doesn't belong there.

madoar: To make this work we will need to add an additional method to the `UiQuestionFactory`. This method needs to be blocking. I think it should have the following signature:

```java
/**
 * Asks the user a question where he/she can answer either yes or no
 * @param questionText The question text
 * @returns True if the user answered yes and false if he answered no
 */
boolean create(String questionText);
```

madoar: @Zemogiter I've just added a new PR for phoenicis that implements the above suggested method.

Can you please try if it works as intended?
I think you will need something like:

```javascript
const answer = uiQuestionFactory.create(
   tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Do you want to continue? Choosing yes will skip DXVK verb installation and continue with other verbs. Choosing no will quit script installation.")
);

// or: `if (answer == false)` 
if (!answer) {
   return this;
}
```

ImperatorS79: I still think there should be three possible answer:
1. Install D9VK anyways (to test MoltenVK)
2. Skip D9VK installation
3. Stop whole app script.

Zemogiter: Thank you. After this change the error is gone and the D9VK installation goes fine.

ImperatorS79: Could you add then in the displayed message that, if they choose to continue, they can still install the verb from the container tab if the want to test it ?

madoar: Can't we support more complex questions via (taken from the League of Legends script):

```
var regions = [
            "EU West",
            "Latin America North",
            "Latin America South",
            "Oceania",
            "Japan",
            "Turkey",
            "Brasil",
            "EU Nordic & East",
            "North America",
            "Russia"
        ];
var selectedRegion = wizard.menu(tr("Select your region:"), regions);
```

Zemogiter: No because no matter whtever the script is executed from other script or via verbs tab they will always get that message.

madoar: Ok, so we basically we need to do this via a popup dialog?

We could also think about adding another method to `uiQuestionFactory`:

```java
String create(String questionText, List<String> choices, String defaultChoice);
```

This method takes a question text and a list with choices. After the user has selected one of the choices the choice text is returned.

Would such a method help?

Zemogiter: @madoar Yeah that sounds more clear than a single yes/no question since we have three options (install D9VK anyways, skip D9VK and go with other verbs and the rest of the script, quit the script alltogether)

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -83,7 +92,7 @@ module.default = class D9VKVerb {
         var wine = new Wine();
         wine.prefix(container);
         var wizard = SetupWizard(InstallationType.VERBS, "D9VK", java.util.Optional.empty());
-        var versions = ["0.12", "0.11", "0.10"];
+        var versions = ["0.13f", "0.13", "0.12", "0.11", "0.10"];
```
madoar: Can you please change all `var` to `const`/`let` in the script? Thanks!

Engines/Wine/Verbs/D9VK/script.js
madoar: @Zemogiter I've just added a new `MultipleChoiceDialog` which can be opened with the previously mentioned command. Can you give it a try?

Please be aware that I didn't test the new dialog. So if it looks kind of strange don't hesitate to tell me so I can fix it (ideally with a screenshot)

Zemogiter: I'm confused as to how the `MultipleChoiceDialog` call should look like.

madoar:
```javascript
const answer = uiQuestionFactory.create(
   tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Do you want to continue? Choosing yes will skip DXVK verb installation and continue with other verbs. Choosing no will quit script installation."),
   ["Choice 1", "Choice 2", "Exit Installer"]
);

if (!answer || answer == "Exit Installer") {
   return this;
}
```

ImperatorS79: it might also be interesting to bypass d9vk installation on computer which does not have enough vulkan support, because wine's d3d9 to OpenGL can to the job.

Zemogiter: Yeah but how we determine vulkan support? Looking at what gpu the user have?

ImperatorS79: Just ask the user if he is sure to install it. I am currently working on a way get such infos.

Zemogiter: OK I'll wait until that PR is merged into POL5.

ImperatorS79:
@Zemogiter there are conflicts.

Zemogiter:
Ready for review.

Zemogiter:
@ImperatorS79 done

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -99,10 +105,9 @@ class D9VK {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "D9VK", Optional.empty());
 
-        const versions = ["0.12", "0.11", "0.10"];
-        const selectedVersion = wizard.menu(tr("Please select the version."), versions, "0.12");
-
-        wine.prefix(container);
+        var wizard = SetupWizard(InstallationType.VERBS, "D9VK", java.util.Optional.empty());
+        const versions = getGithubReleases("Joshua-Ashton", "d9vk", wizard);
+        var selectedVersion = wizard.menu(tr("Please select the version."), versions, versions[0]);
```
madoar: `const`

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -99,10 +105,9 @@ class D9VK {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "D9VK", Optional.empty());
 
-        const versions = ["0.12", "0.11", "0.10"];
-        const selectedVersion = wizard.menu(tr("Please select the version."), versions, "0.12");
-
-        wine.prefix(container);
+        var wizard = SetupWizard(InstallationType.VERBS, "D9VK", java.util.Optional.empty());
```
madoar: `wizard` has already been defined above

Zemogiter:
@madoar done

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -2,17 +2,17 @@ const Wine = include("engines.wine.engine.object");
 const Resource = include("utils.functions.net.resource");
 const { Extractor } = include("utils.functions.filesystem.extract");
 const { ls, cp, remove } = include("utils.functions.filesystem.files");
-
 const operatingSystemFetcher = Bean("operatingSystemFetcher");
-
 const Optional = Java.type("java.util.Optional");
-
 const OverrideDLL = include("engines.wine.plugins.override_dll");
+const uiQuestionFactory = Bean("uiQuestionFactory");
```
madoar: This does not seem to be used anymore? If so please remove the line

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -99,10 +107,8 @@ class D9VK {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "D9VK", Optional.empty());
 
-        const versions = ["0.12", "0.11", "0.10"];
-        const selectedVersion = wizard.menu(tr("Please select the version."), versions, "0.12");
-
-        wine.prefix(container);
```
madoar: Are you sure that this line should be removed?

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -38,22 +37,30 @@ class D9VK {
...
+                return this;
+             }
+             
+             if (answer.text === "NO, quit script alltogether" || !answer) {
+                throw new Error("User aborted the script.");
+             }
+        }
+        else
+        {
+            this.wizard().message(tr("Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else D9VK might not work correctly."));
```
madoar: Please use `wizard` instead of `this.wizard()`

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -38,22 +38,30 @@ class D9VK {
...
-            wizard.message(
-                tr(
-                    "Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else D9VK might not work correctly."
-                )
-            );
+        if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
+        {
+            const answer = wizard.menu(
+                tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation.",
+                ["YES, continue with D9VK installation regardless", "NO, quit script alltogether", "Exit D9VK Installer, but continue with the script"]
```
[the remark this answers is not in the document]
Zemogiter: Shouldn't tr do the trick?

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -2,17 +2,16 @@ const Wine = include("engines.wine.engine.object");
...
 const Optional = Java.type("java.util.Optional");
-
 const OverrideDLL = include("engines.wine.plugins.override_dll");
+const { getGithubReleases } = include("utils.functions.net.githubreleases");
 
 /**
  * Verb to install D9VK
  * see: https://github.com/Joshua-Ashton/d9vk/
  */
+
```
madoar: Please remove this empty line

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -38,22 +37,30 @@ class D9VK {
...
+                return this;
+            }
+
+            if (answer.text === "NO, quit script alltogether" || !answer) {
+                throw new Error("User aborted the script.");
+            }
+        }
+        else
+        {
+            wizard().message(tr("Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else D9VK might not work correctly."));
```
madoar: There is no local function `wizard()`. You need to use `wizard.message(...)`

Engines/Wine/Verbs/D9VK/script.js
```diff
@@ -38,22 +37,30 @@ class D9VK {
...
         }
 
-        if (typeof this.d9vkVersion !== "string") {
-            this.d9vkVersion = "0.12";
+        if (typeof d9vkVersion !== 'string')
```
madoar: `d9vkVersion` is part of the class/object and not a local field inside the `go` method, therefore it must be `this.d9vkVersion`

Zemogiter:
Ready for review

PR #1041, opened 2019-06-28
Update VK9 (fix OS detection)

Zemogiter:
### What works
Everything
### What was not tested
Nothing
### Test
- Operating system (and linux kernel version): Ubuntu 19.04 x64
- Hardware (GPU/CPU):
i7-7700K, GTX 1080 ti
### Ready for review


ImperatorS79:
I do not understand why you want to either skip the verb or the whole installation. The way you did it, this completely prevents someone that would like to test DXVK support under macOS.

Zemogiter:
Why test it if you know it won't work?

ImperatorS79:
Because it might work, providing update to moltenVK, for an older version of DXVK ?

Zemogiter:
Ready for review.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -36,25 +34,26 @@ class VK9 {
...
-            );
-        } else {
-            wizard.message(
-                tr(
-                    "Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else VK9 might not work correctly."
-                )
+        if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
+        {
+            const answer = uiQuestionFactory.create(
+                tr("VK9 is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation."),
```
plata: I think that this decision will be too complicated for many users.

Zemogiter: It's literally just three buttons.

Zemogiter:
Ready for review

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -36,25 +34,28 @@ class VK9 {
...
-            wizard.message(
-                tr(
-                    "Please ensure you have the latest drivers (418.30 minimum for NVIDIA and mesa 19 for AMD) or else VK9 might not work correctly."
-                )
-            );
+        if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
+        {
+            const answer = wizard.menu(
+                tr("VK9 is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation.",
+                ["YES, continue with VK9 installation regardless", "NO, quit script alltogether", "Exit VK9 Installer, but continue with the script"]
```
plata: no translation?

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -1,13 +1,11 @@
...
 const { Extractor } = include("utils.functions.filesystem.extract");
-const { cp, remove } = include("utils.functions.filesystem.files");
-
+const operatingSystemFetcher = Bean("operatingSystemFetcher");
 const Optional = Java.type("java.util.Optional");
-
 const OverrideDLL = include("engines.wine.plugins.override_dll");
-
-const operatingSystemFetcher = Bean("operatingSystemFetcher");
+const uiQuestionFactory = Bean("uiQuestionFactory");
```
madoar: This seems to be unused

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -102,14 +103,11 @@ class VK9 {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "VK9", Optional.empty());
 
-        wine.prefix(container);
+        var wizard = SetupWizard(InstallationType.VERBS, "VK9", java.util.Optional.empty());
```
madoar: `wizard` has already been declared above.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -102,14 +103,11 @@ class VK9 {
...
+        const versions = getGithubReleases("disks86", "VK9", wizard);
+        var selectedVersion = wizard.menu(tr("Please select the version."), versions, versions[0]);
```
madoar: `const`

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -102,14 +103,11 @@ class VK9 {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "VK9", Optional.empty());
 
-        wine.prefix(container);
```
madoar: Are you sure this can be deleted?

Zemogiter: Weird. I swear I did not remove that line.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -101,15 +101,12 @@ class VK9 {
     static install(container) {
         const wine = new Wine();
         const wizard = SetupWizard(InstallationType.VERBS, "VK9", Optional.empty());
-
-        wine.prefix(container);
+        
+        wine.prefix(container); 
+        const versions = getGithubReleases("disks86", "VK9", wizard);
+        const selectedVersion = wizard.menu(tr("Please select the version."), versions, versions[0]);
         wine.wizard(wizard);
```
madoar: Can you please group these statements so that they are semantically more correct:
1) initialize wine:
   ```
   wine.wizard(wizard);
   wine.prefix(container);
   ```
2) afterwards select the verb version:
   ```
   const versions = ...
   const selectedVersion = ....
   ```

This order makes it easier to follow the code flow and it also allows for an easier extension of the verb.

Engines/Wine/Verbs/VK9/script.js
madoar: I agree, @Zemogiter you should use `tr(...)` here to allow translations of the texts. Please be aware that this also requires a change to the `if` conditions below

Engines/Wine/Verbs/VK9/script.js
madoar: A separator line between these two segments is also useful :)

Zemogiter: Done

Zemogiter:
Yes and after it pointed out some errors I fixed them. What do you mean by change to the if's?

madoar:
Your code has a syntax error, `tr(...)` accepts only a single string as its template input.
When working with translations it is better to use something like the following:

```javascript
const question = tr("VK9 is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation.");
const choices = [
   tr("YES, continue with VK9 installation regardless"),
   tr("NO, quit script alltogether"),
   tr("Exit VK9 Installer, but continue with the script")
];

const answer = wizard.menu(question, choices);

switch (answer.index) {
   case 1:
       // choice: "NO, quit script alltogether"
       throw new Error("User aborted the script.");
   case 2: 
       // choice: "Exit VK9 Installer, but continue with the script"
       return this;
   default:
       // do nothing
}
```

Zemogiter:
Ready for review.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -36,25 +33,35 @@ class VK9 {
...
+        if (operatingSystemFetcher.fetchCurrentOperationSystem().getFullName() !== "Linux")
+        {
+            const question = tr("VK9 is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation.")
+            const choices = [
+                tr("YES, continue with VK9 installation regardless"),
+                tr("NO, quit script alltogether"),
+                tr("Exit VK9 Installer, but continue with the script")
+            ];
+
+            const answer = wizard.menu(question, choices);
```
madoar: It may be a good idea to provide a default selection, e.g.:

```javascript
// choose "YES, continue with VK9 installation regardless" by default
const answer = wizard.menu(question, choices, choices[0]);
```

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -36,25 +33,35 @@ class VK9 {
...
+
+            switch (answer.index) {
```
madoar: Just as a note `menu` implementation makes it impossible that `wizard.menu` returns `null`. Therefore this line will always use a valid `index`.

Engines/Wine/Verbs/VK9/script.js
```diff
@@ -36,25 +33,35 @@ class VK9 {
...
+                case 1:
+                    // choice: "NO, quit script alltogether"
+                    throw new Error("User aborted the script.");
+                case 2:
+                    // choice: "Exit VK9 Installer, but continue with the script"
+                    return this;
+                default:
```
madoar: Alternatively to `default:` we could make the `YES, continue with VK9 installation regardless` choice explicit by using:

```javascript
case 0:
   // choice: "YES, continue with VK9 installation regardless"
   break;
```

PR #1044, opened 2019-06-28
Refactor the WineShortcut Class

madoar:
This PR refactors the `WineShortcut` class with the "new" class syntax and some other newer language features.

### DIFF
--- a/Engines/Wine/Shortcuts/Wine/script.js
+++ b/Engines/Wine/Shortcuts/Wine/script.js
@@ -6,6 +6,7 @@ const ShortcutDTOBuilderClass = Java.type("org.phoenicis.library.dto.ShortcutDTO
 /**
  * WineShortcut prototype
  */
+// eslint-disable-next-line no-unused-vars
 class WineShortcut {
     constructor() {
         this._shortcutManager = Bean("shortcutManager");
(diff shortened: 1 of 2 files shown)


Engines/Wine/Shortcuts/Wine/script.js
```diff
@@ -1,186 +1,197 @@
...
-    return this;
-}
+    /**
+     * Sets the miniature for the shortcut
+     *
+     * @param {string[]|URI} miniature An array which specifies the application of which the miniature shall be used or URI of the miniature
+     * @returns {WineShortcut} The WineShortcut object
+     */
+    miniature(miniature) {
+        if (isArray(miniature)) {
```
madoar: @plata @ImperatorS79 @qparis does someone of you know where the `isArray` function is defined? It looks strange to me that we can use a function which we neither defined nor imported...

plata: https://docs.oracle.com/javase/8/docs/jdk/api/nashorn/jdk/nashorn/api/scripting/JSObject.html#isArray--

madoar: Maybe we should try https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray?

plata: Yes.

PR #1045, opened 2019-06-29
Update runtime: fix an error when a directory does not exist

ImperatorS79:
### Description
Only can be seen for a first installation. 
@qparis There is nor runtime x86 with name 201906221529, which will cause bugs since the runtime code expect amd64 and x86 runtime to have the same name (and it should be the same since they are always updated at the same time). I recommand you delete the runtime.json and all the stored runtime archives and properly upload latest amd64 and x86 runtime and a new runtime.json.

### Ready for review

### DIFF
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -169,8 +169,12 @@ var engineImplementation = {
         }
 
         if (download == true) {
-            remove(this._wineEnginesDirectory + "/runtime/lib");
-            remove(this._wineEnginesDirectory + "/runtime/lib64");
+            if(fileExists(this._wineEnginesDirectory + "/runtime/lib")) {
+                remove(this._wineEnginesDirectory + "/runtime/lib");
+            }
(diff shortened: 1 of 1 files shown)


ImperatorS79:
No idea.

qparis:
Same name means that they are build at the exact same date, which is basically impossible. We should allow different names

ImperatorS79:
Why ? do not tell me the name contains the hour and the minute too. If you build them the same day (which should be the case) then there is not a single problem.

qparis:
The name contains the hour and the minute, you can build a runtime twice a day. 
Also, there are some valid reason where one could decide to create a new version of one specific runtime. I see no reason to tie them together

ImperatorS79:
Those are the same runtime for x86 and amd64 architecture. I do not see why there is two archives IMHO. There is very very little chance that you would need to change one and not the other.

qparis:
There are not the same runtime, one of them contains x64 librairies, the other one does not

ImperatorS79:
That is exactly what I said. I could aslo say "Here is **the** runtime for wine". This is true... from a certain point of view ^^.

But fine. I will have to refactor the runtime code :/.

PR #1046, opened 2019-06-30
Refactor the "AppResource" class

madoar:
This PR refactors the `AppResource` class. This contains:
- change the class to use the newer ES 6 `class` syntax
- fetch the `ApplicationDTO` object already in the `application(application)` method
- forbid `null` return values, throw errors instead. Normally a script **expects** a fetched file to exist

### DIFF
--- a/Utils/Functions/Apps/Resources/script.js
+++ b/Utils/Functions/Apps/Resources/script.js
@@ -33,7 +33,7 @@ class AppResource {
         const foundResource = Java.from(this._application.resources).find(resource => resource.name === resourceName);
 
         if (!foundResource) {
-            throw new Error(`Application "${this._application.name}" does not contain resource "resourceName"`);
+            throw new Error(`Application "${this._application.name}" does not contain resource "${resourceName}"`);
         }
 
         return foundResource.content;
--- a/Utils/Functions/Apps/Resources/script.js
+++ b/Utils/Functions/Apps/Resources/script.js
@@ -1,36 +1,40 @@
 /**
-* AppResource prototype
-* @constructor
-*/
-function AppResource() {
-    this._appsManager = Bean("repositoryManager");
-}
-
-/**
-* sets application
-* @param {string} application application of the resource
-* @returns {AppResource} AppResource object
-*/
-AppResource.prototype.application = function (application) {
-    this._application = application;
-    return this;
-}
-
-/**
-* returns resource
-* @param {string} resourceName name of the resource
-* @returns {Resource} found resource
-*/
-AppResource.prototype.get = function (resourceName) {
-    var application = this._appsManager.getApplication(this._application);
-    var foundResource = null;
-    if (application != null && application.resources != null) {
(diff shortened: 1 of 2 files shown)


Utils/Functions/Apps/Resources/script.js
```diff
@@ -1,36 +1,41 @@
...
-    return this;
-}
+    /**
+     * Sets the application containing the resources
+     *
+     * @param {string} application The application with the resource
+     * @returns {AppResource} The AppResource object
+     */
+    application(application) {
+        this._application = this.appsManager.getApplication(application);
```
plata: What if `getApplication` does not work? Does it throw an exception internally?

Utils/Functions/Apps/Resources/script.js
```diff
@@ -1,36 +1,41 @@
...
+     */
+    get(resourceName) {
+        if (!this._application || !this._application.resources) {
+            throw new Error(`Unable to fetch resource from "null" application`);
+        }
+
+        const foundResource = Java.from(this._application.resources).find(resource => resource.name === resourceName);
+
+        if (!foundResource) {
+            throw new Error(`Application "${this._application.name}" does not contain resource "resourceName"`);
```
plata: Shouldn't this print the variable `resourceName`?

Utils/Functions/Apps/Resources/script.js
madoar: What do you mean by does not work? If it returns `null` then https://github.com/PhoenicisOrg/scripts/pull/1046/files#diff-a8f2ab0b620d11b4b57a6e3699a24369R29 should trigger.
If it throws an exception then the exception should be shown inside an `ErrorDialog`

Utils/Functions/Apps/Resources/script.js
madoar: Yes forgot a `$`

madoar: Done

Utils/Functions/Apps/Resources/script.js
plata: Wouldn't it make sense to stop already here when it returns `null` (fail early)?

madoar: I don't know. In theory this is simply a setter, so the script using `AppResource` would still be able to provide a valid application.

plata: Only print a warning then?

madoar: If we assume that the script using the `AppResource` has no access to the created `_application` field I agree, otherwise we shouldn't print a warning in my opinion.

madoar: Do we assume that `_application` is private?

plata: Yes.

madoar: Done

PR #1048, opened 2019-06-30
Refactor the Downloader class

madoar:
This PR refactors the `Downloader` class with the newer `class` syntax

### DIFF
--- a/Utils/Functions/Net/Download/script.js
+++ b/Utils/Functions/Net/Download/script.js
@@ -1,164 +1,183 @@
 include("utils.functions.filesystem.files");
 
 /**
- * Downloader prototype
- * @constructor
+ * Downloader class
  */
-function Downloader() {
-    this._downloader = Bean("downloader");
-    this._algorithm = "SHA";
-    this._headers = {};
-}
-
-/**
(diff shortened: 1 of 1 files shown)


Utils/Functions/Net/Download/script.js
```diff
@@ -1,164 +1,184 @@
...
 
-            if (fileChecksum != this._checksum) {
-                var checksumErrorMessage = tr("Error while calculating checksum for \"{0}\". \n\nExpected = {1}\nActual = {2}",
-                    this._localDestination,
-                    this._checksum,
-                    fileChecksum);
+    /**
+     * Gets the content of the downloaded file
+     *
+     * @returns {String} The content of downloaded file
```
madoar: @qparis @plata I'm not sure if it is a good idea to return file content here. I think this class is normally used to download binary files or other files that don't contain UTF-8 content.

For that reason I think it is better to return the local file path instead. In addition we can thing about renaming the method to `download`

plata: Makes sense but I'm not sure about the impact of this change.

PR #1050, opened 2019-06-30
Fix x86 runtime not being downloaded

ImperatorS79:
### Description
The bug came from the fact that x86 and amd64 runtime does not necessary have the same name
@Kreyren once this is loaded, could delete runtime and runtime.json in .Phoenicis/engines/wine, then retest steam game installation

### Ready for review

### DIFF
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -107,8 +107,10 @@ class WineEngine {
         var runtimeJsonPath = this._wineEnginesDirectory + "/runtime.json";
         var runtimeJson;
         var runtimeJsonFile;
-        var download = false;
-        var name;
+        var downloadx86 = false;
+        var downloadx64 = false;
+        var namex86;
+        var namex64;
         if (!fileExists(runtimeJsonPath)) {
             mkdir(this._wineEnginesDirectory + "/runtime");
             runtimeJsonFile = new Downloader()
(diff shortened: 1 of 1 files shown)


Engines/Wine/Engine/Implementation/script.js
```diff
@@ -199,19 +222,32 @@ class WineEngine {
...
+            });
+            remove(this._wineEnginesDirectory + "/TMP");
+        }
+        if (downloadx86 == true) {
+            if (fileExists(this._wineEnginesDirectory + "/runtime/lib")) {
+                remove(this._wineEnginesDirectory + "/runtime/lib");
+            }
+            mkdir(this._wineEnginesDirectory + "/TMP");
+            var that2 = this;
+            runtimeJson.forEach(function (archive) {
```
madoar: Please use

```javascript
runtimeJson.forEach(archive => {
   ...
});
```

instead. This removes the need to use code like `var that = this`, because you can directly use `this`.

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -107,8 +107,10 @@ class WineEngine {
         var runtimeJsonPath = this._wineEnginesDirectory + "/runtime.json";
         var runtimeJson;
         var runtimeJsonFile;
-        var download = false;
-        var name;
+        var downloadx86 = false;
```
madoar: `let`/`const`?

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -145,73 +155,97 @@ class WineEngine {
...
+        }
+        if (downloadx86 == true) {
+            if (fileExists(this._wineEnginesDirectory + "/runtime/lib")) {
+                remove(this._wineEnginesDirectory + "/runtime/lib");
+            }
+            mkdir(this._wineEnginesDirectory + "/TMP");
+            runtimeJson.forEach(archive => {
+                var runtime;
+                if (archive.name == namex86) {
+                    if (archive.arch == "x86") {
```
madoar: `if (archive.name === namex86 && archive.arch === "x86")`

ImperatorS79:
I do not see currently how to simplify this but this patch is a hotfix so that it works.

ImperatorS79:
@qparis Am I right if I say the greater the archive name, the more up-to-date runtime ? With the current code, it seems phoenicis download an older one :cry: .

ImperatorS79:
@qparis Ok, it seems you did not upload any runtime with the changes in https://github.com/PhoenicisOrg/phoenicis-winebuild/commit/30ee32b7819c2640066fbf1acd88f27054b59091, because I cannot see any freetype libraries in the 201906221528 and 201906221529 runtime. Could you please do so ?

plata:
@ImperatorS79 you should open new issues in winebuild for this kind of problem. Otherwise they might not be noticed.

PR #1057, opened 2019-07-02
Fix: InstallerScript does not wait until installation is finished

plata:


Engines/Wine/QuickScript/Installer Script/script.js
```diff
@@ -72,7 +72,7 @@ class InstallerScript extends QuickScript {
         // back to generic wait (might have been changed in preInstall)
         setupWizard.wait(tr("Please wait..."));
 
-        wine.run(installationCommand.command, installationCommand.args, false, true);
+        wine.run(installationCommand.command, installationCommand.args, null, false, true);
```
madoar: I think it may be a good idea to also add the sixth parameter explicitly: `[]`

plata: If we want to do that, we should do it everywhere.

PR #1064, opened 2019-07-06
Refactor the GOGScript Class

madoar:
This PR refactors the `GOGScript` class

### DIFF
--- a/Engines/Wine/QuickScript/GoG Script/script.js
+++ b/Engines/Wine/QuickScript/GoG Script/script.js
@@ -74,7 +74,7 @@ class GogScript extends QuickScript {
             .wizard(setupWizard)
             .to(tmpDirectory)
             .headers({
-                Authorization: "Bearer " + this._token["access_token"],
+                "Authorization": "Bearer " + this._token["access_token"],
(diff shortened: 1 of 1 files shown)


Engines/Wine/QuickScript/GoG Script/script.js
```diff
@@ -31,54 +33,65 @@ class GogScript extends QuickScript {
...
+        const url = `https://www.gog.com/downloads/${setupFileName}`;
 
         // We set a user agent so that GoG sends the filename of the executable
         return new Downloader()
             .url(url)
             .wizard(setupWizard)
             .to(tmpDirectory)
             .headers({
-                "Authorization": "Bearer " + this._token["access_token"],
+                Authorization: "Bearer " + this._token["access_token"],
```
plata: Why not "" around Authorization anymore?

madoar: Basically because auto-formatting removed it. It is also unnecessary in this case, should I add it again?

plata: I think yes for better understanding (because the next line uses "").

madoar: Done

PR #1074, opened 2019-07-09
Update ZipScript, added a parameter for runing .exe installers inside archives

Zemogiter:
### Description
In it's current shape, the `ZipScript` dosen't do much because there is no information regarding the .exe file inside archive.
### What works
*to be filled*
### What was not tested
Nothing was tested yet, waiting for @ImperatorS79  to test this in his Adobe Photoshop script #1122 
### Test
- Operating system (and linux kernel version): 
- Hardware (GPU/CPU):

### Ready for review


Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -57,7 +57,8 @@ ZipScript.prototype.go = function () {
         .wizard(setupWizard)
         .archive(archive)
         .to(wine.prefixDirectory() + "/drive_c/" + this._name)
-        .extract();
+        .extract()
+        .wine.run();
```
plata: You must pass the extracted executable to `wine.run()`. See e.g. https://github.com/PhoenicisOrg/scripts/blob/749c609a32aace5376853af643b7d8ebbe012921/Engines/Wine/QuickScript/Steam%20Script/script.js#L129

Zemogiter: So it's runInsidePrefix instead of just run?

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -57,7 +57,8 @@ ZipScript.prototype.go = function () {
         .wizard(setupWizard)
         .archive(archive)
         .to(wine.prefixDirectory() + "/drive_c/" + this._name)
-        .extract();
+        .extract()
+        .wine.runInsidePrefix(wine.prefixDirectory() + "/drive_c/" + this._name + ".exe");
```
madoar: I don't think that `wine` is a field of `Extractor`.

Try:

```javascript
   ...
   .extract();

wine.runInsidePrefix(...);
```

instead

Engines/Wine/QuickScript/Zip Script/script.js
plata: Yes.

plata:
Can you maybe test with some other free application so that we are not blocked by Photoshop?

Zemogiter:
I could but I don't know any application other than Photoshop that would make use of ZipScript.

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -57,7 +57,8 @@ ZipScript.prototype.go = function () {
         .wizard(setupWizard)
         .archive(archive)
         .to(wine.prefixDirectory() + "/drive_c/" + this._name)
-        .extract();
+        .extract()
```
madoar: There is a `;` missing after `extract()`. Please also add an empty line between this and `wine.runInside...`. This makes the code better readable.

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -59,6 +59,8 @@ ZipScript.prototype.go = function () {
         .to(wine.prefixDirectory() + "/drive_c/" + this._name)
         .extract();
 
+    wine.runInsidePrefix(wine.prefixDirectory() + "/drive_c/" + this._name + ".exe");
```
madoar: I think this will break most of our existing scripts that make use of `ZipScript`. The reason is that they don't contain an executable in the zip file that should be executed during the installation.

Take [Druid Soccer](https://github.com/PhoenicisOrg/scripts/blob/920c415c38cfffca58d3813108893d1c50c69f44/Applications/Games/Druid%20Soccer/Online/script.js) for example. This installer script only sets an executable that represents the result of the installation but is not the installer. 

In addition I don't think that `this._name` is the name of an `*.exe` file. It normally contains the name of the application. I think we need to add a new field to `ZipScript` that contains the name of the executable in the archive. I think this was also suggested by @plata in https://github.com/PhoenicisOrg/scripts/pull/908#issuecomment-508926377.

ImperatorS79: @Zemogiter Could you test by adding 
```js
	setupPathInsidePrefix(setupPath) {
		this._setupPath = setupPath;
		return this;
	}
```
under url method and
```js
		if (this._setupPath !== 'undefined') {
			wine.runInsidePrefix(this._setupPath);
		}
``` 
inside go method after Extractor ? 
@madoar I think this should do the job.

madoar: Yes, additionally I would recommend adding a `fileExists` check for `this._setupPath` to check that the file really exists before executing it

Zemogiter:
@madoar is this the check sufficient?

madoar:
I think you should invert the check, i.e. add behavior in case the file does not exist. I think ideally we should show an error message informing the user of the problem.

Zemogiter:
@madoar done

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -1,70 +1,79 @@
...
             .wizard(setupWizard)
-            .url(this._url)
-            .checksum(this._checksum)
-            .to(archive)
-            .get();
+            .archive(archive)
+            .to(wine.prefixDirectory() + "/drive_c/" + this._name)
+            .extract();
+        if (!fileExists(this._setupPath)){
+            throw tr("File not found");
```
madoar: Better use:

```javascript
throw new Error(...);
```

Additionally you should add the path to the setup file that couldn't be found to the error message

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -59,11 +60,12 @@ module.default = class ZipScript extends QuickScript {
             .archive(archive)
             .to(wine.prefixDirectory() + "/drive_c/" + this._name)
             .extract();
+        if (fileExists(this._setupPath)){
+	    if (this._setupPath !== 'undefined') {
```
ImperatorS79: The second if is not usefull. If _setupPath is undefined, it will not "exists". You also lost the ability to say to the user that its path is bad if it actually provided one.

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -1,70 +1,76 @@
...
+                .to(archive)
+                .get();
+        }
+
+        new Extractor()
+            .wizard(setupWizard)
+            .archive(archive)
+            .to(wine.prefixDirectory() + "/drive_c/" + this._name)
+            .extract();
+        if (!fileExists(this._setupPath)){
```
ImperatorS79: The problem now: if you not provide a _setupPath (like in regular ZipScript, it will crash. I suggest:
```js
if (this._setupPath !== `undefined`){
    if (!fileExists(this._setupPath)){
        throw "Hello There"
    }
}

**Comment by ImperatorS79** [comment]
I think this is ready to be merged.

**Comment by ImperatorS79** [comment]
I get using this that `the file must be in Phoenicis root directory` (the `_setupPath` is "Adobe Photoshop/Set-up.exe" so that `wine.runInsidePrefix()` launch `drive_c/Adobe Photoshop/Set-Up.exe`)

**Comment by plata** [comment]
Can you show the exact error message? I'm assuming that a prefix is lost at some point.

File: Engines/Wine/QuickScript/Zip Script/script.js [hunk 8]
```diff
@@ -15,6 +17,11 @@ module.default = class ZipScript extends QuickScript {
         return this;
     }
 
+    setupPathInsidePrefix(setupPath) {
```
madoar: Please add comments about:
- how `ZipScript` behaves depending on whether a setup path is specified or not
- what the argument means and what type it is
- that the usage of the method is purely optional

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -58,6 +65,14 @@ module.default = class ZipScript extends QuickScript {
             .to(wine.prefixDirectory() + "/drive_c/" + this._name)
             .extract();
 
+        if (this._setupPath !== `undefined`) {
+            if (!fileExists(this.prefixDirectory() + "/drive_c/" + this._setupPath)) {
+                throw new Error(tr("File \"{0}\" not found", this._setupPath));
+            } else {
```
madoar: The `else` can be removed because `throw` returns in any case

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -1,7 +1,9 @@
 const QuickScript = include("engines.wine.quick_script.quick_script");
 const Downloader = include("utils.functions.net.download");
 const Wine = include("engines.wine.engine.object");
-const { Extractor } = include("utils.functions.filesystem.extract");
+const {Extractor} = include("utils.functions.filesystem.extract");
+const {fileExists} = include("utils.functions.filesystem.files");
+
```
madoar: What is this additional empty line used for?

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -15,6 +16,16 @@ module.default = class ZipScript extends QuickScript {
         return this;
     }
 
+    /**
+     * If setupPath is provided, ZipScript will try to execute the .exe when the archive is extracted 
+     *
+     * @param {string} setupPath path of the setup in <wineprefix>/drive_c/
+     */
```
madoar: `@returns` is missing

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -58,6 +69,13 @@ module.default = class ZipScript extends QuickScript {
             .to(wine.prefixDirectory() + "/drive_c/" + this._name)
```
madoar: This is a bit nitpicking I know but I believe:

```javascript
.to(`${wine.prefixDirectory()}/drive_c/${this._name}`)
```

to be much easier readable

I also think this about the other path strings in this file

Engines/Wine/QuickScript/Zip Script/script.js
ImperatorS79: * @return {ZipScript}
?

madoar: `* @returns {ZipScript} This`

Engines/Wine/QuickScript/Zip Script/script.js
madoar: After this has been fixed I think we can merge this :)

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -1,7 +1,8 @@
 const QuickScript = include("engines.wine.quick_script.quick_script");
 const Downloader = include("utils.functions.net.download");
 const Wine = include("engines.wine.engine.object");
-const { Extractor } = include("utils.functions.filesystem.extract");
+const {Extractor} = include("utils.functions.filesystem.extract");
```
plata: I don't get this: sometimes a space is added around `{}`, sometimes it's removed. How can this happen?

madoar: I think visual studio code normally adds a space for better readability. While other editors seem to remove it

plata: This is nothing the editors should take care of. It should be eslint.

madoar: I this eslint does not care...

madoar: Oh it does: https://eslint.org/docs/rules/object-curly-spacing

madoar: We need a PR where we update the eslint rules:
- https://eslint.org/docs/rules/object-curly-spacing
- the arrow lambda function rule

Engines/Wine/QuickScript/Zip Script/script.js
```diff
@@ -55,9 +67,16 @@ module.default = class ZipScript extends QuickScript {
         new Extractor()
             .wizard(setupWizard)
             .archive(archive)
-            .to(wine.prefixDirectory() + "/drive_c/" + this._name)
+            .to(`${wine.prefixDirectory()}/drive_c/${this._name}`)
             .extract();
 
+        if (this._setupPath !== `undefined`) {
```
ImperatorS79: This check does not seems to work with crayon physics. However the variable is `undefined` so what follows should not be executed :thinking: .

Engines/Wine/QuickScript/Zip Script/script.js
plata: Can you do it?

Engines/Wine/QuickScript/Zip Script/script.js
plata: Should it be
```Javascript
if (this._setupPath !== `undefined`) {
```
or
```Javascript
if (this._setupPath !== undefined) {
```

ImperatorS79: Okay. It works now.

ImperatorS79:
I do not have time to review currently.

Issue #914, opened 2019-04-20
Request: Add FiveM


Kreyren:
![image](https://user-images.githubusercontent.com/11302521/56455259-02fc4280-6354-11e9-89ab-373de0fb16fe.png)

FiveM (https://fivem.net/) is open-source Grand Theft Auto V multiplayer framework for singleplayer.

Relevant: https://github.com/PhoenicisOrg/phoenicis/issues/1919

TODO: Add AppDB entry with test-info.

--- 

Test-Cases wanted!!

plata:
Can you try if simply using staging is sufficient to make this work?

Kreyren:
@plata using quick script with wine-staging is good start and should work based on all provided info.

test cases needed can't test atm, is in my TODO.

Would recommend pushing quick script.

plata:
I do not own GTA V, so I cannot test this myself.

Kreyren:
> I do not own GTA V, so I cannot test this myself.

I sent you the key on stack, go enjoy.

suomilanittaja:
any news?

plata:
According to @Kreyren, `OnlineInstallerScript` with Wine staging should work. Can you try that?

suomilanittaja:
stuck this place using amd64 wine 4.6 staging  https://ibb.co/kB3LvPB

suomilanittaja:
> According to @Kreyren, `OnlineInstallerScript` with Wine staging should work. Can you try that?

i tested wine 4.6 staging (amd64) i get tihs error https://ibb.co/tMkPd8N when i press ok i get other problem which cant be ignored https://ibb.co/tBz8K6X

plata:
Can you try with Windows 10 like suggested there?

suomilanittaja:
i have configured wine to use windows 10

plata:
Mh. In https://bugs.launchpad.net/ubuntu/+source/playonlinux/+bug/1750155 it is suggested to run the exe like
```
wine runas /trustlevel:0x20000 xyz.exe
```
Does this help?

suomilanittaja:
Yes now i can go forward lets wait next problem

plata:
Good. I don't think it's currently possible to do this in a nice way in a quick script. We might have to think about this. Maybe something like `.trustlevel("0x20000")`. Plugin would be nicer but I don't see how this could work currently. Ideas welcome.

suomilanittaja:
[...]

PR #1079, opened 2019-08-10
Fix trust level not being applied

ImperatorS79:
Did not test it but it should be fine I guess.

### DIFF
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -356,11 +356,12 @@ class WineEngine {
         let version = "";
         let architecture = "";
         let workingContainerDirectory = this.getContainerDirectory(this.getWorkingContainer());
+        let distribution;
 
         if (fileExists(workingContainerDirectory)) {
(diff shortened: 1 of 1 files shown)


Engines/Wine/Engine/Implementation/script.js
```diff
@@ -356,11 +356,12 @@ class WineEngine {
         let version = "";
         let architecture = "";
         let workingContainerDirectory = this.getContainerDirectory(this.getWorkingContainer());
+        let distribution;
```
madoar: `let distribution;` or `let distribution = "";` (e.g. similar to `version` and `architecture`)

plata:
Shouldn't it be "trust level" in the title?

madoar:
Is there an easy way for us to test this?

plata:
I think it's used by an application.

madoar:
I don't think it is used yet. As far as I could find out it (trustLevel) seems to have been added because of #914

plata:
I see. Then I'm not sure how to test.

madoar:
@plata I think it is fine to merge this, we lose nothing if we do

madoar:
@plata you decide

PR #1088, opened 2019-08-17
Refactor the "d3dx9" Verb Script

madoar:
This PR modernizes the `d3dx9` verb script


Engines/Wine/Verbs/d3dx9/script.js
```diff
@@ -1,6 +1,6 @@
 const Wine = include("engines.wine.engine.object");
 const Resource = include("utils.functions.net.resource");
-const {CabExtract} = include("utils.functions.filesystem.extract");
+const { CabExtract } = include("utils.functions.filesystem.extract");
```
plata: Why isn't this handled by the formatter?

madoar: I have the habit of pressing the formatting keyboard shortcut directly before pressing the save shortcut. I think that this change comes from that.

PR #1100, opened 2019-08-23
Improve Code Quality of the Utils Scripts

madoar:
This PR improves the code quality of the `Utils` scripts by:
- extracting the `Bean(...)` and `Java.type(...)` calls to the front of the utils scripts
- removing the now unnecessary `// eslint-disable-next-line no-unused-vars` comments

In addition this PR fixes some small issues

### DIFF
--- a/Utils/Functions/Filesystem/Files/script.js
+++ b/Utils/Functions/Filesystem/Files/script.js
@@ -1,4 +1,5 @@
-var fileUtilities = Bean("fileUtilities");
+const fileUtilities = Bean("fileUtilities");
+const checksumCalculator = Bean("checksumCalculator");
 
 /**
  * Lists all files and directories contained in the given path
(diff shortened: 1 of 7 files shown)


Utils/Functions/Filesystem/Files/script.js
```diff
@@ -211,29 +210,15 @@ module.Checksum = class Checksum {
      * @returns {string} The calculated checksum
      */
     get() {
+        let progressBar;
         if (this._wizard) {
-            var progressBar = this._wizard.progressBar(tr("Checking file consistency..."));
```
madoar: The previous call led to `progressBar` being available only inside the if block but not afterwards

Utils/Functions/Net/Download/script.js
```diff
@@ -119,19 +119,20 @@ module.default = class Downloader {
             this._message = tr("Please wait while {0} is downloaded...", this._fetchFileNameFromUrl(this._url));
         }
 
+        let progressBar;
         if (this._wizard) {
-            var progressBar = this._wizard.progressBar(this._message);
```
madoar: The previous call led to `progressBar` being available only inside the if block but not afterwards

Issue #1098, opened 2019-08-23
Steam does not work


Zemogiter:
When using a SteamScript, the exe process is displayed and you get the usual "connecting to the steam account" messages. But after that the .exe shuts itself down and nothing happens. Tested using `Space Engineers` prefix with all upstream wine versions between 4.0.1 and 4.14 same results
```
wine: cannot find L"C:\\windows\\system32\\winemenubuilder.exe"
000b:err:wineboot:process_run_key Error running cmd L"C:\\windows\\system32\\winemenubuilder.exe -a -r" (2).
001b:err:setupapi:SetupDiSelectBestCompatDrv No compatible drivers were enumerated for device L"WINEMOUSE\\VID_0000&PID_0000\\0&WINEMOUSE&0&0".
001b:err:plugplay:install_device_driver Install function 0x17 failed, error 0xe0000228.
0039:err:mscoree:LoadLibraryShim error reading registry key for installroot
0039:err:mscoree:LoadLibraryShim error reading registry key for installroot
0039:err:mscoree:LoadLibraryShim error reading registry key for installroot
0039:err:mscoree:LoadLibraryShim error reading registry key for installroot
0039:err:msi:execute_command unable to execute command 2
0025:err:module:load_so_dll failed to load .so lib "/home/jonasz/.Phoenicis/engines/wine/upstream-linux-amd64/4.14/bin/../lib64/wine/winegstreamer.dll.so": /lib/x86_64-linux-gnu/libgstaudio-1.0.so.0: undefined symbol: gst_aggregator_simple_get_next_time
0055:err:module:load_so_dll failed to load .so lib "/home/jonasz/.Phoenicis/engines/wine/upstream-linux-amd64/4.14/bin/../lib/wine/winegstreamer.dll.so": /lib/i386-linux-gnu/libgstaudio-1.0.so.0: undefined symbol: gst_aggregator_simple_get_next_time
wine: configuration in '/home/jonasz/.Phoenicis/containers//wineprefix//Space Engineers' has been updated.
[0823/111145:ERROR:network_change_notifier_win.cc(170)] WSALookupServiceBegin failed with: 122
008a:err:module:load_so_dll failed to load .so lib "/home/jonasz/.Phoenicis/engines/wine/upstream-linux-amd64/4.14/bin/../lib/wine/winepulse.drv.so": libpulsecommon-10.0.so: nie można otworzyć pliku obiektu dzielonego: Nie ma takiego pliku ani katalogu
[...]

PR #1110, opened 2019-08-25
Update Space Engineers, bumped up wine version and fixed few typos

Zemogiter:
### Description
It should be `no-cef-sandbox` not `no-ces-sandbox`
### What works
Everything but sometimes Wine decides to force ALSA instead of PulseAudio. This results in even choppier audio. But when I use winetricks to set it to pulse and go to the sound tab in winecfg I get "none" next to the "chosen driver".
### What was not tested
Nothing I can think of at this moment
### Test
- Operating system (and linux kernel version): Ubuntu 19.04 x64 5.0.0-25-generic
- Hardware (GPU/CPU):
i7-7700K, GTX 1080 ti
### Ready for review

### DIFF
--- a/Applications/Games/Space Engineers/Steam/script.js
+++ b/Applications/Games/Space Engineers/Steam/script.js
@@ -9,8 +9,8 @@ new SteamScript()
     .name("Space Engineers")
     .editor("Keen Software House")
     .author("Zemogiter")
-    .appId(244850)
-    .wineVersion("4.11")
+    .appId("244850")
+    .wineVersion("4.14")
     .wineDistribution("upstream")
     .wineArchitecture("amd64")
     .preInstall(function (wine, wizard) {
@@ -24,4 +24,4 @@ new SteamScript()
         wizard.message(tr("You have to install libjpeg62 and libjpeg62-dev or else the thumbnails in New Game menu will be dispalyed as magenta rectangles."));
         wizard.message(tr("Due to JIT compiler issues and the way this game uses multithreating, there are audio stutters. This script will attempt to minimize them but you might also have to enter the alsoft-conf command in terminal and set sample depth to 32bit float and period size to 2048."));
     })
-    .executable("Steam.exe", ["-silent", "-applaunch", 244850, "-no-ces-sandbox", "-skipintro"])
+    .executable("Steam.exe", ["-silent", "-applaunch", "244850", "-no-cef-sandbox", "-skipintro"])


Applications/Games/Space Engineers/Steam/script.js
```diff
@@ -9,8 +9,8 @@ new SteamScript()
     .name("Space Engineers")
     .editor("Keen Software House")
     .author("Zemogiter")
-    .appId(244850)
-    .wineVersion("4.11")
+    .appId("244850")
```
madoar: @plata should the `appId` be passed as a string or as a number?

plata: Does it matter?

madoar: I don't think so but I think we should do it in a uniform manner.

plata: Isn't it uniform in the other Steam scripts?

madoar: According to other scripts `appId` should be a number, see:

https://github.com/PhoenicisOrg/scripts/blob/920c415c38cfffca58d3813108893d1c50c69f44/Applications/Games/Enderal/Steam/script.js#L8

https://github.com/PhoenicisOrg/scripts/blob/920c415c38cfffca58d3813108893d1c50c69f44/Applications/Games/BRINK/Steam/script.js#L7

https://github.com/PhoenicisOrg/scripts/blob/920c415c38cfffca58d3813108893d1c50c69f44/Applications/Games/BioShock/Steam/script.js#L7

Maybe we should enforce this and throw an error if some invalid input has been given?

Zemogiter: Well I have to renistall the prefix anyways because Wine stuck with ALSA again and any attempt at changing it to pulse results in what I have written in OP. Might as well check if passing `appId` as a string matters.

plata: > Maybe we should enforce this and throw an error if some invalid input has been given?

Yes.

madoar: This leaves us with the initial question:

> Should the `appId` be passed as a string or as a number?

Zemogiter: When installing `Space Engineers` I get the same error from https://github.com/PhoenicisOrg/scripts/issues/1098#issuecomment-525303874

plata: It should be number.

Zemogiter: Changing `appId` back to number changes nothing. Still getting that error.

madoar: @Zemogiter I think you misunderstand the comment discussions. This is not about the error you are getting. We simply want to have a uniform style how the `appId` should be passed to the `SteamScript`.

Zemogiter: @madoar understood

PR #1117, opened 2019-09-06
Remove superfluous this.install() and this._wizard call

ImperatorS79:
### Description
If we get there the this.install has already been called

### Ready for review


plata:
I don't think that this will work. If the second `install()` is removed, you will not be able to change the Wine version for an existing prefix (iirc).

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -120,7 +120,7 @@ module.default = class WineEngine {
             mkdir(this._wineEnginesDirectory + "/runtime");
 
             runtimeJsonFile = new Downloader()
-                .wizard(this._wizard)
+                .wizard(setupWizard)
```
plata: I can't see where this `setupWizard` comes from. Is it simply `undefined`?

ImperatorS79: setupWizard is passed to the function in `this.install` using `this.getWizard`

plata: Didn't see that. Actually: what's the point of doing that?

madoar: @plata we have a similar situation with our pre- and post-hooks in the installers. These hooks take both a `wine` parameter and a `wizard` parameter. The `wizard` is kind of redundant because you can fetch it using `wine.wizard()`.

ImperatorS79: It was just for uniformizing. I can replace all of those call with `this.getWizard()`.

plata: I think we should remove the `wizard` parameter if it is not required (both in this file and for the hooks).

ImperatorS79: I removed the setupWizard in the function prototype.

madoar: @ImperatorS79 do you also want to provide a PR where the wizard argument is removed from the installer scripts?

ImperatorS79: This is a lot of work...

madoar: If you don't have the time that is fine. I couldn't refrain from asking ;)
Alternatively we need an issue where we are reminded about this

plata:
The change is fine for me. I'm thinking if we shouldn't change
```Javascript
if (fileExists(workingContainerDirectory)) {
    const containerConfiguration = configFactory.open(workingContainerDirectory + "/phoenicis.cfg");

    distribution = containerConfiguration.readValue("wineDistribution", "upstream");
    architecture = containerConfiguration.readValue("wineArchitecture", "x86");

    const operatingSystem = operatingSystemFetcher.fetchCurrentOperationSystem().getWinePackage();

    subCategory = distribution + "-" + operatingSystem + "-" + architecture;
    version = containerConfiguration.readValue("wineVersion");

    this.install(subCategory, version);
} else {
    print('Wine prefix "' + this.getWorkingContainer() + '" does not exist!');

    return "";
}
```
to
```Javascript
if (!fileExists(workingContainerDirectory)) {
    print('Wine prefix "' + this.getWorkingContainer() + '" does not exist!');

    return "";
}

const containerConfiguration = configFactory.open(workingContainerDirectory + "/phoenicis.cfg");

distribution = containerConfiguration.readValue("wineDistribution", "upstream");
architecture = containerConfiguration.readValue("wineArchitecture", "x86");

const operatingSystem = operatingSystemFetcher.fetchCurrentOperationSystem().getWinePackage();

subCategory = distribution + "-" + operatingSystem + "-" + architecture;
version = containerConfiguration.readValue("wineVersion");

this.install(subCategory, version);
```
Might be clearer. What do you think?

madoar:
```javascript
if (!fileExists(workingContainerDirectory)) {
    print('Wine prefix "' + this.getWorkingContainer() + '" does not exist!');

    return "";
}
```

Should we maybe throw an error/exception instead if this case occurs?

plata:
Yes.

plata:
There are issues in Codacy.

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -365,23 +367,21 @@ module.default = class WineEngine {
         let workingContainerDirectory = this.getContainerDirectory(this.getWorkingContainer());
         let distribution = "";
 
-        if (fileExists(workingContainerDirectory)) {
-            const containerConfiguration = configFactory.open(workingContainerDirectory + "/phoenicis.cfg");
+        if (!fileExists(workingContainerDirectory)) {
+            throw tr('Wine prefix "{0}" does not exist', this.getWorkingContainer());
```
madoar: `throw new Error(...);`

madoar:
@ImperatorS79 this PR has a merge conflict.

Engines/Wine/Engine/Implementation/script.js
madoar: After the `new` has been added I think the PR can be merged

plata:
@madoar you have requested changes. Have they been addressed? If so, can we merge this?

madoar:
There is still the `new` in `throw new Error(...);` missing.

plata:
@ImperatorS79 can you fix that?

plata:
@madoar please review again.

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -368,23 +368,21 @@ module.default = class WineEngine {
         let workingContainerDirectory = this.getContainerDirectory(this.getWorkingContainer());
         let distribution = "";
 
-        if (fileExists(workingContainerDirectory)) {
-            const containerConfiguration = configFactory.open(workingContainerDirectory + "/phoenicis.cfg");
+        if (!fileExists(workingContainerDirectory)) {
+            throw new tr('Wine prefix "{0}" does not exist', this.getWorkingContainer());
```
madoar: It needs to be: `throw new Error(tr(...));`

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -368,23 +368,21 @@ module.default = class WineEngine {
...
+        }
 
-            distribution = containerConfiguration.readValue("wineDistribution", "upstream");
-            architecture = containerConfiguration.readValue("wineArchitecture", "x86");
+        const containerConfiguration = configFactory.open(workingContainerDirectory + "/phoenicis.cfg");
 
-            const operatingSystem = operatingSystemFetcher.fetchCurrentOperationSystem().getWinePackage();
+        distribution = containerConfiguration.readValue("wineDistribution", "upstream");
+        architecture = containerConfiguration.readValue("wineArchitecture", "x86");
```
madoar: Is it still necessary to declare these values above with `let` or could we declare them here using `const`?

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -368,23 +368,21 @@ module.default = class WineEngine {
...
 
-            subCategory = distribution + "-" + operatingSystem + "-" + architecture;
-            version = containerConfiguration.readValue("wineVersion");
+        const operatingSystem = operatingSystemFetcher.fetchCurrentOperationSystem().getWinePackage();
 
-            this.install(subCategory, version);
-        } else {
-            print('Wine prefix "' + this.getWorkingContainer() + '" does not exist!');
+        subCategory = distribution + "-" + operatingSystem + "-" + architecture;
+        version = containerConfiguration.readValue("wineVersion");
```
madoar: Same as above

PR #1123, opened 2019-09-11
Avoid that runtime is installed multiple times during one installation

### LINKED ISSUE #1055: "Download runtime json..." displayed as default message while installing a script
See #959 at the end.

plata:
It is sufficient to download/update the runtime once during one installation. We can safely assume that the runtime does not change during this period of time.

fixes #1055

### DIFF
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -26,6 +26,7 @@ module.default = class WineEngine {
         this._wineWebServiceUrl = propertyReader.getProperty("webservice.wine.url");
         this._wizard = null;
         this._workingContainer = "";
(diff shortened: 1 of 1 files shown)


ImperatorS79:
You always have to fetch the json from internet to check if the runtime has been updated. This PR just avoids that if the runtime json have already been fetched in one installation it is not fetched again during this installation.

ImperatorS79:
I do not understand what you mean here.

madoar:
We can use the approach I described above in all situations, independent of whether we are calling the same method multiple times during one installation or across multiple installations. We don't need a second "fallback" solution.

ImperatorS79:
I still do not understand why what has been done is a hack. We do not have the checksum of each file inside the runtime, and it would be insane to check everything.

PR #1131, opened 2019-09-27
Change Plugins to use the Module System

madoar:
This PR changes the plugins to use the module system.

### DIFF
--- a/Engines/Wine/Plugins/Windows version/script.js
+++ b/Engines/Wine/Plugins/Windows version/script.js
@@ -1,72 +1,69 @@
-const Wine = include("engines.wine.engine.object");
-
-include("engines.wine.plugins.regedit");
+const Regedit = include("engines.wine.plugins.regedit");
 
 /**
- * default windows version
- * @param {string} version (win7, vista, win2003, winxp, win2k, winnt, winme, win98, win95, win31)
- * @param {string} [servicePack] e.g. sp3
- * @returns {string|Wine} get: Windows version, set: Wine object
+ * Plugin to set the windows version
  */
-Wine.prototype.windowsVersion = function (version, servicePack) {
-    // get
-    if (arguments.length == 0) {
-        return this.regedit().fetchValue(["HKEY_CURRENT_USER", "Software", "Wine", "Version"]);
+module.default = class WindowsVersion {
+    constructor(wine) {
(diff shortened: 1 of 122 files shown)


madoar:
@ImperatorS79 @Zemogiter can you please check if the changes are working as intended
@plata if it is ok I would like to add a documentation for the plugins in a separate issue. As far as I remember we have no documentation for plugins yet so it is takes more effort compared to simply fixing an existing documentation :)

Travis fails because it can't build the github pages @plata any ideas?

Engines/Wine/Plugins/Windows version/script.js
```diff
@@ -1,72 +1,69 @@
...
-        "REGEDIT4\n" +
-        "\n";
+        return this;
+    }
 
-    that.wine = function (wine) {
-        that._wine = wine;
-        return that;
-    };
+    go() {
```
madoar: There are some plugins which were previously two separate `Wine` functions. One of these is the `WindowsVersion` plugin which previously consisted of:
- `windowsVersion` and
- `setOsForApplication`

I've tried to unify all such cases into a single plugin. This may lead to some strange behavior compared to before because the new behavior may differ if a plugin is used twice, once to set the global windows version and then a second time to set a local application-specific windows version

plata:
I'm fine to update the documentation separately. Regarding the GitHub pages: https://community.netlify.com/t/jekyll-builds-consistently-failing-with-error-superclass-must-be-a-class-faraday-deprecatedconstant-given/3799/7.

madoar:
It seems like a restart of travis fixed the issue :)
Any change requests?

plata:
There was another update to faraday. Probably that fixed it.

Zemogiter:
I think the old way we defined the dll override was more readable.

madoar:
@Zemogiter I agree the previous way how we accessed verbs and plugins was more readable. The issue was just that it was not as safe to use as it is now. For example different plugins and verbs overrode the same functions of the `Wine` class. This can lead to very unexpected behavior. In addition we had problems with missing includes or unused includes. The new verb and plugin implementations fix these issues.

Maybe it is important to note that we can still improve the implementations. While we know that they are not perfect we may find solutions to the identified issues and improve on them later!

PR #1133, opened 2019-10-02
Add property to (not) use Wine runtime

plata:
set `application.environment.wineRuntime`, by default the runtime is used

fixes https://github.com/PhoenicisOrg/phoenicis/issues/1940

### DIFF
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -447,12 +447,17 @@ module.default = class WineEngine {
             ldPath = userData.ldPath + ldPath;
         }
 
+        let runtimePath = "";
+        let runtimePath64 = "";
+        if (this._useRuntime) {
+            runtimePath = this._wineEnginesDirectory + "/runtime/lib/";
+            runtimePath64 = this._wineEnginesDirectory + "/runtime/lib64/";
+        }
+
         if (architecture == "amd64") {
             ldPath =
-                this._wineEnginesDirectory +
-                (this._useRuntime ? "/runtime/lib64/:" : "") +
-                this._wineEnginesDirectory +
-                (this._useRuntime ? "/runtime/lib/:" : "") +
+                runtimePath64 + ":" +
+                runtimePath + ":" +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib64/:" +
                 this.getLocalDirectory(subCategory, version) +
@@ -460,8 +465,7 @@ module.default = class WineEngine {
                 ldPath;
         } else {
             ldPath =
-                this._wineEnginesDirectory +
-                (this._useRuntime ? "/runtime/lib/:" : "") +
+                runtimePath + ":" +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib/:" +
                 ldPath;
--- a/Engines/Wine/Engine/Implementation/script.js
+++ b/Engines/Wine/Engine/Implementation/script.js
@@ -23,6 +23,7 @@ module.default = class WineEngine {
         this._ldPath = propertyReader.getProperty("application.environment.ld");
         this._wineEnginesDirectory = propertyReader.getProperty("application.user.engines") + "/wine";
         this._winePrefixesDirectory = propertyReader.getProperty("application.user.containers") + "/" + WINE_PREFIX_DIR + "/";
+        this._useRuntime = (propertyReader.getProperty("application.environment.wineRuntime") !== "false");
         this._wineWebServiceUrl = propertyReader.getProperty("webservice.wine.url");
         this._wizard = null;
         this._workingContainer = "";
@@ -43,7 +44,9 @@ module.default = class WineEngine {
     }
 
     install(subCategory, version) {
-        this._installRuntime(this.getWizard());
+        if (this._useRuntime) {
+            this._installRuntime(this.getWizard());
+        }
 
         const [distribution, , architecture] = subCategory.split("-");
         const localDirectory = this.getLocalDirectory(subCategory, version);
@@ -447,9 +450,9 @@ module.default = class WineEngine {
         if (architecture == "amd64") {
             ldPath =
                 this._wineEnginesDirectory +
-                "/runtime/lib64/:" +
+                (this._useRuntime ? "/runtime/lib64/:" : "") +
                 this._wineEnginesDirectory +
-                "/runtime/lib/:" +
+                (this._useRuntime ? "/runtime/lib/:" : "") +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib64/:" +
                 this.getLocalDirectory(subCategory, version) +
@@ -458,7 +461,7 @@ module.default = class WineEngine {
         } else {
             ldPath =
                 this._wineEnginesDirectory +
-                "/runtime/lib/:" +
+                (this._useRuntime ? "/runtime/lib/:" : "") +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib/:" +
                 ldPath;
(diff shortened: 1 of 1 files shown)


Engines/Wine/Engine/Implementation/script.js
```diff
@@ -23,6 +23,7 @@ module.default = class WineEngine {
         this._ldPath = propertyReader.getProperty("application.environment.ld");
         this._wineEnginesDirectory = propertyReader.getProperty("application.user.engines") + "/wine";
         this._winePrefixesDirectory = propertyReader.getProperty("application.user.containers") + "/" + WINE_PREFIX_DIR + "/";
+        this._useRuntime = (propertyReader.getProperty("application.environment.wineRuntime") !== "false");
```
madoar: Is it possible to set this property from inside Phoenicis?

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -447,9 +450,9 @@ module.default = class WineEngine {
         if (architecture == "amd64") {
             ldPath =
                 this._wineEnginesDirectory +
-                "/runtime/lib64/:" +
+                (this._useRuntime ? "/runtime/lib64/:" : "") +
```
madoar: I think an `if (this._useRuntime) { ... } else { ... }` may be easier to read and understand

ImperatorS79: It is perfectly readable. There is not a lot of ternary operator ^^.

Engines/Wine/Engine/Implementation/script.js
plata: No. It would be a bit problematic to do that because Phoenicis should be agnostic of the Engine and this property is specific for Wine.

Engines/Wine/Engine/Implementation/script.js
plata: If I use `if else`, I cannot do it inline. So I would have to repeat much of the code which is not readable either.

madoar: Actually I'm against an inline solution. I think a single string would be the most readable, e.g.:
```javascript
ldPath = `${this._wineEnginesDirectory}/runtime/lib/:/runtime/lib/:${this.getLocalDirectory(subCategory, version)}/lib/:${ldPath}`;
```

madoar: An alternative would be to define the parts of `ldPath` in separate variables and concat them at the end:

```javscript
ldPath = `${phoenicisRuntime}:${systemRuntime}:${localLibraries}:${ldPath}`;
```

plata: Better?

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -444,21 +447,25 @@ module.default = class WineEngine {
             ldPath = userData.ldPath + ldPath;
         }
 
+        let runtimePath = "";
+        let runtimePath64 = "";
+        if (this._useRuntime) {
+            runtimePath = this._wineEnginesDirectory + "/runtime/lib/";
+            runtimePath64 = this._wineEnginesDirectory + "/runtime/lib64/";
+        }
```
madoar: Does it make sense to move this to another internal/private method? Something like:

```javascript
fetchRuntimePaths() {
   if (this._useRuntime) {
      return {
         runtimePath: `{this._wineEnginesDirectory}/runtime/lib/`,
         runtimePath64: `{this._wineEnginesDirectory}/runtime/lib64/`
      };
   } else {
      return { runtimePath: "", runtimePath64: "" };
   }
}
```

Afterwards we can fetch the runtime paths with:

```javascript
const { runtimePath, runtimePath64 } = this.fetchRuntimePaths();
```

I think this looks more clean compared to a `let` solution

Engines/Wine/Engine/Implementation/script.js
```diff
@@ -444,21 +447,25 @@ module.default = class WineEngine {
...
-                this._wineEnginesDirectory +
-                "/runtime/lib64/:" +
-                this._wineEnginesDirectory +
-                "/runtime/lib/:" +
+                runtimePath64 + ":" +
+                runtimePath + ":" +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib64/:" +
                 this.getLocalDirectory(subCategory, version) +
                 "/lib/:" +
```
madoar: Can we move these two paths also to variables?

Engines/Wine/Engine/Implementation/script.js
madoar: Ok

Engines/Wine/Engine/Implementation/script.js
plata: I don't like that. My problem with this solution is that you would naturally assume that the returned path can be used where ever the runtime path is required (e.g. when extracting the runtime archive). However, that's not the case. You can only use it in the context of LD path.

Engines/Wine/Engine/Implementation/script.js
plata: We could but honestly I do not see the advantage of it. Also it's not related to this PR at all.

Engines/Wine/Engine/Implementation/script.js
madoar: We could also define `fetchRuntimePaths` as an internal method (e.g. inside the method where it is actually used):

```javascript
const fetchRuntime = () => {
   if (this._useRuntime) {
      return {
         runtimePath: `{this._wineEnginesDirectory}/runtime/lib/`,
         runtimePath64: `{this._wineEnginesDirectory}/runtime/lib64/`
      };
   } else {
      return { runtimePath: "", runtimePath64: "" };
   }
};
```

Would this solve the problem in your opinion?

Engines/Wine/Engine/Implementation/script.js
madoar: It makes the composition of the final `ldPath` string easier to read

Engines/Wine/Engine/Implementation/script.js
plata: Yes it would but I still believe that we should not do it:
1. I don't really see the difference to the current implementation.
2. In general we should avoid using JS possibilities which might not be easily understandable for people who do not know JS (unless it gives us a big advantage).

Engines/Wine/Engine/Implementation/script.js
plata: Done.

Engines/Wine/Engine/Implementation/script.js
madoar: The braces are not required

plata: I know. I've put them only to make it more clear.

Engines/Wine/Engine/Implementation/script.js
madoar: > I don't really see the difference to the current implementation.

The main difference is that it removes `let`

> In general we should avoid using JS possibilities which might not be easily understandable for people who do not know JS (unless it gives us a big advantage).

What exactly do you mean by JS possibilities? The "inner function" or the destructuring? Both are features that are also supported by a lot of other programming languages. Inner functions for example are also supported by Java:

```java
Runnable fetchRuntime = () => {
   ...
};
```

plata: I meant the inner function. We have to keep in mind that the people who get in contact with the scripts might know nothing about programming. The simpler we keep our scripts, the easier they will be understandable for them.

madoar: I think certain parts of the scripts may be public but are normally used as blackbox components by the script users/developers. I count the engine implementations in this category. These classes/modules should be documented extensively so that the typical user won't have questions about what a method does.

plata: That's a valid point.

PR #1134, opened 2019-10-02
Replace hardcoded version list by git-fetched version list

ImperatorS79:
### Description
The versions can be fetched from github so that we do not have to update the scripts when new versions are out. Thanks  @Kreyren for the infos.

### What works
Currently it fails saying getGitVersions is not a function. @plata @madoar.

### Ready for review


plata:
Should it be `const { getGitVersions }`?

ImperatorS79:
Ok working now.

Utils/Functions/Net/GitVersions/script.js
```diff
@@ -0,0 +1,35 @@
...
+        .message(tr("Fetching versions list..."))
+        .to(tmpDir + "/releases.json")
+        .get();
+
+    let versions = new Array();
+
+    const releasesFileJson = JSON.parse(cat(releasesFile));
+    releasesFileJson.forEach(version => {
+        versions.push(version.tag_name);
+    });
```
madoar:
```javascript
const versions = JSON.parse(cat(releasesFile)).map(version => version.tag_name);
```

?

Engines/Wine/Verbs/gallium9/script.js
```diff
@@ -34,7 +36,8 @@ class Gallium9 {
         const system32directory = this.wine.system32directory();
 
         if (typeof this.gallium9Version !== "string") {
-            this.gallium9Version = "0.4";
+            const versions = getGitVersions("iXit", "wine-nine-standalone", wizard);
+            this.gallium9Version = versions[0];
```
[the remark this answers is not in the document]
ImperatorS79: The first element should always be the latest release so yes.

Utils/Functions/Net/GitVersions/script.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Get an array of tag corresponding to the version in a git repository
+ *
+ * @param {string} repositoryOwner owner of the repository
+ * @param {string} repositoryName name of the repository
+ * @param {wizard} wizard the wizard
+ * @returns {string[]} list of version tag in the repository
+ */
+function getGitVersions(repositoryOwner, repositoryName, wizard) {
```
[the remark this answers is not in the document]
ImperatorS79: I do not like the split. I do not see why we should split them.

ImperatorS79:
Gallium9 output this error:
```js
[ERROR] org.phoenicis.multithreading.ControlledThreadPoolExecutorService (l.64) - -64
        at org.phoenicis.win32.pe.PEReader.readRealModeStubProgram(PEReader.java:85)
        at org.phoenicis.win32.pe.PEReader.parseExecutable(PEReader.java:31)
        at org.phoenicis.tools.win32.ExeAnalyser.isArchitecture(ExeAnalyser.java:62)
        at org.phoenicis.tools.win32.ExeAnalyser.is64Bits(ExeAnalyser.java:46)
        at <js> run(Unnamed:410:15290-15336)
        at <js> run(Unnamed:171:5412-5502)
        at <js> go(Unnamed:90:3297-3376)
        at <js> install(Unnamed:111:3948-4004)
        at org.graalvm.sdk/org.graalvm.polyglot.Value.invokeMember(Value.java:457)
        at org.phoenicis.engines.VerbsManager.lambda$installVerb$0(VerbsManager.java:71)
        at org.phoenicis.scripts.session.PhoenicisInteractiveScriptSession.eval(PhoenicisInteractiveScriptSession.java:35)
        at org.phoenicis.scripts.interpreter.BackgroundScriptInterpreter.lambda$createInteractiveSession$1(BackgroundScriptInterpreter.java:45)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:835)
Caused by host exception: java.lang.NegativeArraySizeException: -64
```
Does anyone have this (you will need this PR because it has a fix for a bug in the path of the exe) ?

Utils/Functions/Net/GitVersions/script.js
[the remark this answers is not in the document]
plata: I'd suggest one function `getGitHubReleases`.

Utils/Functions/Net/GithubReleases/script.js
```diff
@@ -27,4 +27,4 @@ function getGithubVersions(repositoryOwner, repositoryName, wizard) {
 
     return versions;
 }
-module.getGithubVersions = getGithubVersions;
+module.getGithubReleases = getGithubReleases;
```
madoar: If the method does not call itself (recursively) it is ok to directly export it, i.e.:

```javascript
module.getGithubReleases = function getGithubReleases(repositoryOwner, repositoryName, wizard) {
   ...
};
```

Engines/Wine/Verbs/gallium9/script.js
```diff
@@ -2,6 +2,9 @@ const Wine = include("engines.wine.engine.object");
 const Resource = include("utils.functions.net.resource");
 const { Extractor } = include("utils.functions.filesystem.extract");
 const { remove, lns } = include("utils.functions.filesystem.files");
+const { getGithubReleases } = include("utils.functions.net.githubreleases");
+
```
madoar: I think you can remove one of the two empty lines. One should be enough :)

Utils/Functions/Net/GithubReleases/script.js
ImperatorS79: I did this like in the other utils function (ls, remove, ...)

madoar: Yes I split it there in two parts (the function definition and the `module` assignment) because `touch` uses `fileExists` which does not work in the same module if `fileExists` is directly exported. (You would need to use `module.fileExists()` in this case which is much harder to read)

Utils/Functions/Net/GitVersions/script.js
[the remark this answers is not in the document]
madoar: After taking a more detailed look at the code it also makes no sense to split the method, because we're talking here about a Github specific API.

PR #1137, opened 2019-10-20
Update Wine LATEST_DEVELOPMENT_VERSION and LATEST_STAGING_VERSION to 4.18

plata:
merge when available


Zemogiter:
How come the x32 branches have the wine 4.18 already compiled while x64 branches does not?

qparis:
There is an issue with x64 builds.

Investigating on it

Zemogiter:
Thank you @qparis we can merge this now.

Issue #1138, opened 2019-11-03
Steam no longer support windows xp after the library update


Zemogiter:
When Steam is done downloading updates and `SteamScript` tries to start Steram.exe for the first time you get the message that says Steam no longer supports your os. This is a problem because now we can no longer use the worarrounds to make steam browser work. I've tried it on Lutris since it applies none of the workarrounds we are using and the Library tab now works like steam browser (store, community etc.). Which is to say it dosen't work at all. 
This situation makes every steam script broken until we figure out a solution. Applies to both new `SteamScript` installations and existing prefixes.

plata:
I have no idea how we could solve this. What I do not understand: Steam apps are installed/started via Steam api. This should work even without the library.

Zemogiter:
We just need to make the library tab work because now it's either black or blinking.

plata:
What will happen if we do not set it to winxp? If Steam does not run with winxp anymore anyways, I think we can safely remove the overrides.

Zemogiter:
> What will happen if we do not set it to winxp?

The steam will launch thus allowing user to login and close to let the script continue. The issue is the steam library tab either not working at all (black page) or partially working (page displayed with black rectangles or the page turns to black when scrolling).

plata:
Ok, but like I said this should not be required to install/play games. So let's do it.

Zemogiter:
Closed via #1140

PR #1140, opened 2019-11-09
Update SteamScript (remove winxp overrides)

Zemogiter:
### Description
After Steam overhauled their Library tab, support for windows xp was dropped entirely. This results in Steam client unable to start which breaks the sequence of events neceseary to get this script to work. Still need to figure out a new way to make both steam browser and library tab work. Currently both elements either dosen't work at all (black page) or work partially (page goes to black when scrolling, page displayed with black rectangles on some content, makes the `<div>` elements visible). A partial fix to #1138 
### What works
Installing games.
### What was not tested
Nothing
### Test
- Operating system (and linux kernel version): Ubuntu 19.10 5.3.0-19-generic
- Hardware (GPU/CPU): i7-7700K, GTX 1080 ti

### Ready for review

### DIFF
--- a/Applications/Games/Steam/Online/script.js
+++ b/Applications/Games/Steam/Online/script.js
@@ -2,7 +2,6 @@ const OnlineInstallerScript = include("engines.wine.quick_script.online_installe
 const { LATEST_STAGING_VERSION } = include("engines.wine.engine.versions");
 
 const Corefonts = include("engines.wine.verbs.corefonts");
-const WindowsVersion = include("engines.wine.plugins.windows_version");
 
 new OnlineInstallerScript()
     .name("Steam")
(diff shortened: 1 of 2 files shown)


plata:
This should also be changed in https://github.com/PhoenicisOrg/scripts/blob/master/Applications/Games/Steam/Online/script.js.

Applications/Games/Steam/Online/script.js
```diff
@@ -2,7 +2,6 @@ const OnlineInstallerScript = include("engines.wine.quick_script.online_installe
 const { LATEST_STAGING_VERSION } = include("engines.wine.engine.versions");
 
 const Corefonts = include("engines.wine.verbs.corefonts");
-const WindowsVersion = include("engines.wine.plugins.windows_version");
```
plata: I think this include can be removed in the SteamScript as well.

Zemogiter: No because there is an uplay fix that uses this include.

PR #1149, opened 2019-12-06
Add Mf workaround

Zemogiter:
### Description
A workaround for Media Foundation
https://github.com/z0z0z/mf-install
Needed for games that rely on it for ingame video and cutscenes. For example, benchmark videos in PC Building Simulator. After following instructions in the repo linked above, I can report they work as on Windows now.
### What works
Extracting the file, copying the dll files
### What does not work
dll override adds "native" to the name of the overriden dll not where it supoused to be, regedit dosen't generate any error but I see no registry entry from mf.reg
### What was not tested
Everything after above mentioned 
### Test
- Operating system (and linux kernel version): Ubuntu 19.10
- Hardware (GPU/CPU):i7-7700K,GTX 1080ti

### Ready for review
- [ ] `json-align` and `eslint` run according to the [documentation](https://phoenicisorg.github.io/scripts/General/tools/). 
- [ ] Codacy and travis checked.


Zemogiter:
Output of eslint:
```
  79:68    error  Unnecessary escape character: \S  no-useless-escape
  79:77    error  Unnecessary escape character: \W  no-useless-escape
  79:82    error  Unnecessary escape character: \L  no-useless-escape
  79:576   error  Unnecessary escape character: \C  no-useless-escape
  79:582   error  Unnecessary escape character: \{  no-useless-escape
  79:675   error  Unnecessary escape character: \C  no-useless-escape
  79:681   error  Unnecessary escape character: \{  no-useless-escape
  79:720   error  Unnecessary escape character: \I  no-useless-escape
  79:795   error  Unnecessary escape character: \C  no-useless-escape
  79:801   error  Unnecessary escape character: \{  no-useless-escape
  79:888   error  Unnecessary escape character: \C  no-useless-escape
  79:894   error  Unnecessary escape character: \{  no-useless-escape
  79:933   error  Unnecessary escape character: \I  no-useless-escape
  79:1008  error  Unnecessary escape character: \C  no-useless-escape
  79:1014  error  Unnecessary escape character: \{  no-useless-escape
  79:1107  error  Unnecessary escape character: \C  no-useless-escape
  79:1113  error  Unnecessary escape character: \{  no-useless-escape
  79:1152  error  Unnecessary escape character: \I  no-useless-escape
```

plata:
Why does this PR contain Applications/Games/Heroes of Might & Magic IV/miniatures/main.png?

Zemogiter:
I don't know. I've created separate branches for both Heroes 4 and this PR from master.

plata:
Can you fix it?

Zemogiter:
Done

Zemogiter:
These escape characters are needed.

Zemogiter:
OP was updated

Zemogiter:
Tried referencing the .reg file inside `mfWorkaroundDir` but the same issue happens.

madoar:
By issue you mean that the .reg file is not modified?

madoar:
@Zemogiter `patch` requires two parameters to work, please see:

https://github.com/PhoenicisOrg/scripts/blob/93403d446bc8f0068eddd968fdfe99282df94fb5/Engines/Wine/Plugins/regedit/script.js#L39-L48

Zemogiter:
By issue I mean the registry entries from mf.reg are not in registry editor.

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,113 @@
...
+                if (file.endsWith(".dll")) {
+                    cp(`${mfWorkaroundDir}/system32/${file}`, system64directory);
+
+                    new OverrideDLL(this.wine).withMode("native", [file]).go();
+                }
+            });
+        }
+
+        const regeditFile = `${prefixDirectory}/mfWorkaround/mf-install-master/mf.reg`;
+        new Regedit(this.wine).patch(regeditFile);
```
madoar: Try `new Regedit(this.wine).open(regeditFile);`

Zemogiter: Tried that. It did not fix the issue.

madoar: `patch` is definitely wrong. See my argument [above](https://github.com/PhoenicisOrg/scripts/pull/1149#issuecomment-569492163).

What "error" do you get when you use `open`?

Zemogiter: Thats the thing. No error in terminal window in either `patch` or `open`.

plata: One parameter should work (the type can be `undefined`). I think it should be
```Javascript
new Regedit(this.wine).patch(cat(regeditFile));
```
`Regedit#patch` writes the `patchContent` to a file. So in your case, it tries to patch the registry with the string "${prefixDirectory}/mfWorkaround/mf-install-master/mf.reg" instead of the content of that file.

madoar: Yes this is basically the goal. For now I think it is more important to get it building with the least amount of effort. For now I think we should simply [ignore](http://junit.sourceforge.net/javadoc/org/junit/Ignore.html) the test or remove it completely. Most likely I will not be able to do this today.

madoar: I don't see the difference to using `open`? The point is that the initial start consists of a regedit file which we want to `open` with regedit. `patch` takes a regedit content string, writes it to a temporary file and opens it at the end. This is basically `open` with an additional previous I/O call

Zemogiter: @plata this is weird. The part of .reg file that was meant to be in `HKEY_LOCAL_MACHINE` is nowhere in registry editor but everything after that is there.

plata: @madoar you are right, this should be the same as `open`.
@Zemogiter do you get the same result if you run regedit manually (on the command line)?

Zemogiter: @madoar yes

plata: In this case it cannot be a Phoenicis issue. Maybe something is wrong in the file?

Zemogiter: @plata just tried it on plain Wine and the whole .reg file is in the registry editor.

madoar: Have you tried other scripts that make use of `Regedit`? Do they work for you?

Zemogiter: @madoar yes they do

plata: Strange. If Phoenicis use the same command line as plain Wine, how can it be different...

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,113 @@
...
+const Optional = Java.type("java.util.Optional");
+const Regedit = include("engines.wine.plugins.regedit");
+const Regsvr32 = include("engines.wine.plugins.regsvr32");
+const OverrideDLL = include("engines.wine.plugins.override_dll");
+
+/**
+ * Verb to install Media Foundation workaround
+ * see: https://github.com/z0z0z/mf-install
+ */
+class mfWorkaround {
```
madoar: Classes are normally named using title case, i.e. `MFWorkaround`

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,113 @@
...
+     * Sets the used mfWorkaround version
+     *
+     * @param {string} mfWorkaroundVersion The version of mfWorkaround to downlaod
+     * @returns {mfWorkaround} The mfWorkaround object
+     */
+    withVersion(mfWorkaroundVersion) {
+        this.mfWorkaroundVersion = mfWorkaroundVersion;
+
+        return this;
+    }
```
madoar: If `mfWorkaroundVersion` can't be used I would recommend removing the field and the method for now because it is more confusing that useful

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,113 @@
...
+        const wizard = this.wine.wizard();
+        const prefixDirectory = this.wine.prefixDirectory();
+        const system32directory = this.wine.system32directory();
+        const system64directory = this.wine.system64directory();
+
+        //commented out because it's useless until z0z0z will use releases
+        /*if (typeof this.mfWorkaroundVersion !== "string") {
+            const versions = getGithubReleases("Kron4ek", "FAudio-Builds", wizard);
+            this.mfWorkaroundVersion = versions[0];
+        }*/
```
madoar: The same applies to these comments

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,113 @@
...
+
+        const mfWorkaroundDir = `${prefixDirectory}/mfWorkaround/mf-install-master`;
+
+        ls(`${mfWorkaroundDir}/syswow64`).forEach(file => {
+            if (file.endsWith(".dll")) {
+                cp(`${mfWorkaroundDir}/syswow64/${file}`, system32directory);
+
+                new OverrideDLL(this.wine).withMode("native", [file]).go();
+            }
+        });
```
madoar: A valid alternative could be:

```javascript
ls(`${mfWorkaroundDir}/syswow64`)
   .filter(file => file.endsWith(".dll"))
   .forEach(file => {
      cp(`${mfWorkaroundDir}/syswow64/${file}`, system32directory);

      new OverrideDLL(this.wine).withMode("native", [file]).go();
   });
```

Engines/Wine/Verbs/mfWorkaround/script.js
Zemogiter: I've left them because someone may notice gthub link and ask why we are not using `getGithubReleases` function.

Engines/Wine/Verbs/mfWorkaround/script.js
Zemogiter: Why is the current option bad?

madoar: It is not bad per se. I just wanted to show you an alternative style that you can you but don't need to. This comment is purely optional

Engines/Wine/Verbs/mfWorkaround/script.js
madoar: @plata what do you think should we retain these comments or remove them?

Engines/Wine/Verbs/mfWorkaround/script.js
```diff
@@ -0,0 +1,108 @@
...
+    constructor(wine) {
+        this.wine = wine;
+    }
+
+    /**
+     * Sets the used mfWorkaround version
+     *
+     * @param {string} mfWorkaroundVersion The version of mfWorkaround to downlaod
+     * @returns {mfWorkaround} The mfWorkaround object
+     */
```
madoar: You forgot to remove this comment block

Engines/Wine/Verbs/mfWorkaround/script.js
plata: Remove. If @Zemogiter would like to have a comment, use something like:
```
// TODO: use GitHub releases if repository provides releases
```
We should not keep commented code like this. It will likely get outdated and only pollutes the script.

Engines/Wine/Verbs/mfWorkaround/script.js
Zemogiter: Understood

madoar:
Has the regedit problem be solved or is it still work in progress?

Zemogiter:
@madoar still work in progress

Issue #1151, opened 2019-12-12
Check Plugin Scripts


madoar:
After the last changes to the plugin scripts the structure of many plugins has changed. Therefore we are not sure at the moment which plugin scripts are currently in a working state.

This issue will function as a summary of the testing progress of the plugin scripts. It contains a list of all implemented plugins with a checkbox that should be ticked if the plugin script has been tested and it is working as intended. If it is not working the checkbox will be replaced with a comment containing what is not working so that we can fix the plugin.

If some plugins are missing feel free to add a comment with the missing verbs and I will add them to the list

- [ ] DOS support
- [ ] DirectDraw renderer
- [ ] Font smoothing
- [ ] GLSL
- [ ] OpenGL version
- [ ] UseTakeFocus
- [ ] Windows version 
- [ ] csmt 
- [ ] hdpi 
- [ ] managed 
- [ ] native application
- [ ] nocrashdialog 
- [ ] override DLL 
- [ ] regedit 
- [ ] register font 
- [ ] regsvr32 
- [ ] sound driver 
- [ ] virtual desktop

plata:
What is the definition of "works"? Is it sufficient if it installs without crashing?

Zemogiter:
There is something wrong with override DLL plugin. The overrides are placed right next to the names of the dlls. There are three questionmarks between `()` meaning wine does not recorgnize. Also only the first dll override mentioned in ` new OverrideDLL(this.wine).withMode("native, builtin", ["atl100", "msvcp100", "msvcr100", "vcomp100"]).go();` is placed in "Libraries" tab in winecfg. Tested with `vcrun2010` and `mfc42`.

plata:
I'm not sure I understand completely what you wrote. Can you describe step by step what you did and what you observed?

Zemogiter:
When trying to install `Space Engineers` I've discovered this:
![Zrzut ekranu z 2019-12-17 19-09-23](https://user-images.githubusercontent.com/24385408/71022446-cb22e600-2100-11ea-9170-4e896d21ba64.png)
Then I've quited the installation and tried installing each verb separatley to figure out what's wrong. Every verb gave me these question marks because for some reason `override DLL` plugin started to place override modes (like `disabled`, `native` etc.) not inside the brackets `()` but between them and the dll name.

plata:
Can you check how it looks like in the registry?

Zemogiter:
[...]

PR #1152, opened 2019-12-23
Fix Wine plugin "Override DLL"

plata:
see #1151

### DIFF
--- a/Engines/Wine/Plugins/override DLL/script.js
+++ b/Engines/Wine/Plugins/override DLL/script.js
@@ -3,11 +3,11 @@ const Regedit = include("engines.wine.plugins.regedit");
 module.default = class OverrideDLL {
     constructor(wine) {
         this.wine = wine;
-        this.modes = {};
+        this.modes = new Map();
     }
 
     withMode(mode, libraries) {
-        this.modes[mode] = libraries;
+        this.modes.set(mode, libraries);
 
         return this;
     }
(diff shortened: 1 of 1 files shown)


plata:
@Zemogiter please try.

Engines/Wine/Plugins/override DLL/script.js
```diff
@@ -3,23 +3,21 @@ const Regedit = include("engines.wine.plugins.regedit");
 module.default = class OverrideDLL {
     constructor(wine) {
         this.wine = wine;
-        this.modes = {};
+        this.modes = new Map();
```
madoar: Indeed a `Map` is a better choice here.

Engines/Wine/Plugins/override DLL/script.js
```diff
@@ -3,23 +3,21 @@ const Regedit = include("engines.wine.plugins.regedit");
...
+        this.modes.set(mode, libraries);
 
         return this;
     }
 
     go() {
         let regeditFileContent = `REGEDIT4\n\n[HKEY_CURRENT_USER\\Software\\Wine\\DllOverrides]\n`;
 
-        Object.entries(this.modes)
-            .map(([mode, libraries]) => libraries.map(library => [library, mode]))
```
madoar: Just as a sidenot, I think the following change would be enough to fix the issue but I agree that your code is more readable:

```javascript
.flatMap(([mode, libraries]) => libraries.map(library => [library, mode]))
```

Zemogiter:
@plata thanks now it works

PR #1153, opened 2019-12-24
Update latest Wine versions automatically

### LINKED ISSUE #445: update latest Wine version automatically
Currently we update LATEST_DEVELOPMENT_VERSION and LATEST_STAGING_VERSION manually. It should be possible to automate this, i.e. update whenever a new version is available.

plata:
### Description
fixes #445

### What works
general concept

### What was not tested
all the scripts


Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,122 @@
...
+const {cat, touch} = include("utils.functions.filesystem.files");
+const Downloader = include("utils.functions.net.download");
+const propertyReader = Bean("propertyReader");
+
+/**
+ * Sorts an array of Wine versions
+ * @param {array} versions The versions array
+ * @returns {void}
+ */
+module.sortVersions = function (versions) {
```
madoar: Please add that you are doing inplace sorting

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,122 @@
...
+    versions.sort((a, b) =>
+    {
+        const aVersionNumbers = Array.from(a.split('.')).map(item => Number(item));
+        const bVersionNumbers = Array.from(b.split('.')).map(item => Number(item));
```
madoar: Maybe it makes sense to first check that the version strings `a` and `b` have the correct format, e.g. consist of three number parts split by dots (a regex could be used for this).

Afterwards we should be able to "simplify" the sorting code:

```javascript
const [aMajor, aMinor, aPatch] = Array.from(a.split('.')).map(item => Number(item));
const [bMajor, bMinor, bPatch] = Array.from(b.split('.')).map(item => Number(item));

if (aMajor != bMajor) {
   return aMajor - bMajor;
}

if (aMinor != bMinor) {
   return aMinor - bMinor;
}

return aPatch - bPatch;
```

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,122 @@
...
+
+
+module.getLatestStableVersion = function (wizard) {
+    const versionsJson = module.getAvailableVersions(wizard);
+
+    const packages = versionsJson
+        .filter(distribution => distribution.name === "upstream-linux-x86")
+        .flatMap(distribution => distribution.packages);
+
+    const regex = new RegExp('^\\d+\\.0(\\.\\d+)?$');
```
madoar: There is a dedicated regex syntax in JS that we can use here. This also removes the need to escape the `\` by using: `\\`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

Engines/Wine/Engine/Versions/script.js
madoar: If this function is only used internally inside this module we should make it "module private", i.e. not add it to the exports.

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,122 @@
...
+        .flatMap(distribution => distribution.packages);
+
+    const regex = new RegExp('^\\d+\\.\\d(\\.\\d+)?$');
+    const develVersions = packages
+        .filter(({version}) => regex.test(version))
+        .flatMap(package => package.version);
+
+    module.sortVersions(develVersions);
+
+    return develVersions[develVersions.length-1];
```
madoar: It looks to me as if we have this code block trice, each time with only a single string change. Can we move this to a generic "module private" method that accepts a `wizard` and a `distributionName` input?

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,122 @@
...
+}
+
+module.getLatestStagingVersion = function (wizard) {
```
madoar: `get...` or `fetch...`?

madoar:
To make the `wizard` accessible inside the installer script we could think about passing `getLatestStagingVersion` as a function to `wineVersion` by calling:

```javascript
   ...
   .wineVersion(getLatestStagingVersion)
   ...
```

This would allow us to "inject" the `wizard` from inside the implementation of `wineVersion`

plata:
I've thought about that as well but I don't really like it. It requires that we call the function in every QuickScript and it's a bit strange to pass a function here but a value e.g. for `wineArchitectur`. I have no better solution though.

madoar:
> it's a bit strange to pass a function here but a value e.g. for wineArchitecture

We could do the same for `wineArchitecture`

> I've thought about that as well but I don't really like it. It requires that we call the function in every QuickScript

I like the style of passing functions where it makes sense. From a user perspective it is also quite clear to me what happens. It only becomes problematic if an error occurs for some reason and we provide no useful error messages.

plata:
I'm really would like to avoid that people have to write
```Javascript
.wineVersion(function() { return "4.0.3"; } );
```
instead of
```Javascript
.wineVersion("4.0.3");
```

madoar:
You don't have this problem. We can allow **both** a function input and a string input. This is possible if you check the type of parameter in the `wineVersion` method.

Engines/Wine/Engine/Versions/script.js
plata: The string doesn't have to be three numbers split by dots. You can have "4.0" or "5.0.0-rc2".

plata:
I will try.

Engines/Wine/Engine/Versions/script.js
plata: It's also `getAvailableVersions`.

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,139 @@
...
+        const aVersionParts = a.version.match(versionRegExp);
+        const bVersionParts = b.version.match(versionRegExp);
+
+        const aVersionNumbers = Array.from(aVersionParts[1].split('.')).map(item => Number(item));
+        const bVersionNumbers = Array.from(bVersionParts[1].split('.')).map(item => Number(item));
+
+        // ensure that cases where not all version numbers (major, minor, patch) are set are handled correctly
+        const maxVersionIdx = 2;
+        let i;
+        for (i = 0; i <= maxVersionIdx; i++) {
```
madoar: you can move the `let` inside the for loop, i.e.:

```javascript
for (let i = 0; i <= maxVersionIdx; i++) {
```

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,139 @@
...
+            if (typeof aVersionNumbers[i] === 'undefined') {
```
madoar: or `if (!aVersionNumbers[i])` ?

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,139 @@
...
+        }
+    }
+    );
+}
+
+/**
+ * Fetches the latest available Wine version based on the category and version regex
+ * @param {wizard} wizard setup wizard to show the download progress of the versions json
+ * @param {string} category (e.g. "upstream-linux-x86")
+ * @param {string} regex regex the version must fulfill
```
madoar: `regex` is not a `string` but a `RexExp` object I think

Engines/Wine/QuickScript/Quick Script/script.js
```diff
@@ -135,6 +135,20 @@ module.default = class QuickScript {
...
     }
 
+    /**
+     * determines which Wine version should be used
+     * required in case the version is computed by a function
+     * @param {wizard} wizard setup wizard (e.g. to show download progress of versions json)
+     * @returns {void}
+     */
+    _determineWineVersion(wizard) {
+        if (this._wineVersion && {}.toString.call(this._wineVersion) === '[object Function]') {
```
madoar: alternatively `if (this._wineVersion && typeof this._wineVersion === "function")` and `if (this._wineVersion && this._wineVersion instanceof Function)` should also work

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,139 @@
...
+    const packages = versionsJson
+        .filter(distribution => distribution.name === category)
+        .flatMap(distribution => distribution.packages);
+
+    const regExp = new RegExp(regex);
+    const versions = packages
+        .filter(({version}) => regExp.test(version))
+        .flatMap(package => package.version);
+
+    return versions[versions.length-1];
```
madoar: Is it always the case that the last version is the "latest" version?

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,139 @@
...
+    const versionsJson = JSON.parse(cat(versionsFile));
+
+    versionsJson.forEach(distribution => sortVersions(distribution.packages));
+
+    return versionsJson;
+}
+
+
+module.getLatestStableVersion = function (wizard) {
+    return getLatestVersion(wizard, "upstream-linux-x86", /^\d+\.0(\.\d+)?$/);
```
madoar: Should we forward the `category` too?

Engines/Wine/QuickScript/Quick Script/script.js
```diff
@@ -135,6 +135,20 @@ module.default = class QuickScript {
         return this;
     }
 
+    /**
+     * determines which Wine version should be used
+     * required in case the version is computed by a function
+     * @param {wizard} wizard setup wizard (e.g. to show download progress of versions json)
+     * @returns {void}
+     */
+    _determineWineVersion(wizard) {
```
madoar: I don't like that this method modifies `this._wineVersion`. I think it should return the extracted wine version instead of overriding `this._wineVersion`.

Engines/Wine/Engine/Versions/script.js
madoar: We could rename that :)

Engines/Wine/QuickScript/Quick Script/script.js
plata: Then you can never use `this._wineVersion` anywhere in the QuickScripts. I don't think that's very intuitive. I agree that the current handling is a bit strange but I don't like your solution either.

Engines/Wine/Engine/Versions/script.js
plata: Do we want that? It would enter the code if `aVersionNumbers[i]` is 0.

Engines/Wine/QuickScript/Quick Script/script.js
plata: I've used the accepted answer of this StackOverflow: https://stackoverflow.com/questions/5999998/check-if-a-variable-is-of-function-type. I didn't really check why this is the accepted answer or why the other possibilities have not been accepted, but honestly, I don't see a reason to change it.

Engines/Wine/Engine/Versions/script.js
plata: I think so. Do you have any particular use case in mind?

Engines/Wine/Engine/Versions/script.js
plata: When calling `getLatestStableVersion`? I don't think so. What I've thought about is: we could pass the Wine architecture. Normally x86 and amd64 should be updated at the same time. What might maybe happen: the x86 version has just been built but the amd64 not yet. I will add this.

Engines/Wine/Engine/Versions/script.js
plata: But not now. I'm fine if we decide in general when to use `get` and when `fetch` and then apply this everywhere later.

Engines/Wine/QuickScript/Quick Script/script.js
madoar: Why?

My suggestion is to introduce a local variable `selectedWineVersion` or `resolvedWineVersion` that is used in the QuickScripts. This should not be that difficult because you are calling `_determineWineVersion` in the QuickScripts anyway right?

To better illustrate what I mean:

```javascript
const resolvedWineVersion = this._determineWineVersion(wizard);

// do something with resolvedWineVersion instead of this._wineVersion
```

Engines/Wine/Engine/Versions/script.js
madoar: I don't have an use case in mind. The issue is just that this code allows us in theory to process engine version lists from multiple sources. Over some of these sources we have control, there we can ensure the ordering of the versions. Other sources we possibly know of but have no control. In addition there can also be a number of sources that we possibly don't even know of. 

For this reason I think we should add some additional logic or rules to ensure that the "latest" version is returned.

Engines/Wine/QuickScript/Quick Script/script.js
madoar: You can use the solution you have. It is just not that intuitive in my opinion. I honestly don't even know what `{}.toString.call(this._wineVersion)` does...

My alternatives in comparison are easier (I would even say almost trivial) to understand

Engines/Wine/Engine/Versions/script.js
madoar: You are right.

Just asking but does `aVersionNumbers[i] == undefined` work?

plata: I've checked the sorted list and it looked ok.

Engines/Wine/Engine/Versions/script.js
plata: I don't see how we could determine the latest version if this logic does not work.

Engines/Wine/QuickScript/Quick Script/script.js
plata: I can do that.

Engines/Wine/Engine/Versions/script.js
madoar: We can sort them before we take the first/last entry right?

```javascript 
const versions = ...

sortVersions(versions);

return versions[versions.length - 1];
```

Engines/Wine/Engine/Versions/script.js
```diff
@@ -1,8 +1,138 @@
...
+    const versionsJson = module.getAvailableVersions(wizard);
+
+    const packages = versionsJson
+        .filter(distribution => distribution.name === category)
+        .flatMap(distribution => distribution.packages);
+
+    const regExp = new RegExp(regex);
+    const versions = packages
+        .filter(({version}) => regExp.test(version))
+        .flatMap(package => package.version);
```
madoar: `.flatMap(package => package.version)` or `.map(package => package.version)`?

Engines/Wine/QuickScript/Quick Script/script.js
plata: I've changed it.

Engines/Wine/Engine/Versions/script.js
plata: This is what I do already. The `getAvailableVersions` does the sorting.

Engines/Wine/Engine/Versions/script.js
plata: I'm not 100% sure (I'm not that familiar with this stuff). In any case it's working.

Engines/Wine/Engine/Versions/script.js
plata: Just realized that you are referring to `==` vs `===`. I have no clue. I would assume it should work as well.

madoar: I'm not sure if `===` works as well therefore I wanted to be sure and wrote `==` :D

Engines/Wine/Engine/Versions/script.js
madoar: Then ignore this comment :)

Engines/Wine/Engine/Versions/script.js
plata: Well, the current implementation works. Everything else I can't say with certainty ;).

Engines/Wine/Engine/Versions/script.js
madoar: `map` applies a function on a given list/array and returns a list/array with the results. E.g. if you do the following 

```javascript
[1,2,3].map(i => [i, i])
```

you will get `[[1,1], [2,2], [3,3]]`

`flatMap` also applies a function on a given list/array and returns a list/array with the results, but differing from `map` the function needs to return a list/array and the final list/array will be "flattened". This will lead to a different result for:

```javascript
[1,2,3].flatMap(i => [i, i])
```

you will get `[1,1, 2,2, 3,3]`

So the main question is: does `package.version` contain an array or a simple string?

plata: It's a simple string. From my understanding, `map` and `flatMap` should behave the same way in this case.

madoar: In Java it would not work. You would get a compile time error, because the API expects the function to return a `Stream` (i.e. a collection) and not a single value.

Maybe it is a particularity of Javascript?

plata: I'm not sure. Just tried `map` and it works as well. I will change it.

Engines/Wine/QuickScript/Quick Script/script.js
```diff
@@ -90,7 +90,11 @@ module.default = class QuickScript {
     }
 
     wineVersion(wineVersion) {
-        this._wineVersion = wineVersion;
+        if (wineVersion && wineVersion instanceof Function) {
+            this._wineVersionFunction = wineVersion;
+        } else {
+            this._wineVersionFunction = function () { return wineVersion; };
```
madoar: `function() {return wineVersion; }` or `() => wineVersion` ?

plata: I think it's purely a question of personal preference. Let's go with the current implementation.

PR #1087, opened 2019-08-17
Refactor the "vcrun2003" Verb Script

madoar:
This PR refactors the `vcrun2003` verb script and fixes #1086

PR #1156, opened 2019-12-28
ESLint: add object-curly-spacing rule

plata:


madoar:
Can you also add the rule for the arrow lambda function syntax? (see https://github.com/PhoenicisOrg/scripts/pull/1087#discussion_r314975839)

plata:
see #1157

PR #1168, opened 2020-01-03
Add "Automatic Rebase" GitHub action

plata:
see https://github.com/marketplace/actions/automatic-rebase

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,0 +1,22 @@
+on: 
+  issue_comment:
+    types: [created]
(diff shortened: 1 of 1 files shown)


madoar:
Would this GitHub action work for us (because of: https://github.com/cirrus-actions/rebase/issues/12)?

plata:
Mh. Didn't see that.

madoar:
@plata how should we continue with this PR?

plata:
Doesn't seem like there will be a solution anytime soon so we can as well close this.

PR #1169, opened 2020-01-03
ESLint action test: break Wine script

plata:


plata:
Any ideas why the action is not working?

Zemogiter:
Login error?

plata:
I've deleted the action again because it seems that it only works if the repository is built with npm/yarn.

madoar:
@plata can this PR be closed?

PR #1144, opened 2019-12-01
Add Heroes of Might and Magic IV

Zemogiter:
### Description
When the wineprefix is created the installation crashes wiht this error:
```
[ERROR] org.phoenicis.multithreading.ControlledThreadPoolExecutorService (l.64) - Input/Output error
	at java.base/java.io.FileInputStream.readBytes(Native Method)
	at java.base/java.io.FileInputStream.read(FileInputStream.java:257)
	at org.apache.commons.io.input.ProxyInputStream.read(ProxyInputStream.java:78)
	at org.phoenicis.win32.pe.PEReader.readDosHeader(PEReader.java:93)
	at org.phoenicis.win32.pe.PEReader.parseExecutable(PEReader.java:30)
	at org.phoenicis.tools.win32.ExeAnalyser.isArchitecture(ExeAnalyser.java:62)
	at org.phoenicis.tools.win32.ExeAnalyser.is64Bits(ExeAnalyser.java:46)
	at <js> run(Unnamed:413:15444-15490)
	at <js> run(Unnamed:171:5412-5502)
	at <js> go(Unnamed:88:3412-3493)
	at com.oracle.truffle.polyglot.ObjectProxyHandler.invoke(HostInteropReflect.java:678)
	at com.sun.proxy.$Proxy66.go(Unknown Source)
	at org.phoenicis.javafx.components.application.skin.ApplicationInformationPanelSkin.lambda$installScript$7(ApplicationInformationPanelSkin.java:237)
	at org.phoenicis.scripts.session.PhoenicisInteractiveScriptSession.eval(PhoenicisInteractiveScriptSession.java:35)
	at org.phoenicis.scripts.interpreter.BackgroundScriptInterpreter.lambda$createInteractiveSession$1(BackgroundScriptInterpreter.java:45)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by host exception: java.io.IOException: Błąd wejścia/wyjścia

[WARNING] 
org.graalvm.polyglot.PolyglotException: Input/Output error
    at java.io.FileInputStream.readBytes (Native Method)
    at java.io.FileInputStream.read (FileInputStream.java:257)
    at org.apache.commons.io.input.ProxyInputStream.read (ProxyInputStream.java:78)
[...]

PR #1170, opened 2020-01-05
Update Installer Script (run postInstall before shortcut creation)

Zemogiter:
### Description
Thanks to #1144 we have discovered that scripts will fail if `.executable()` points to a file that only exist after `postInstall` step is done. Since @plata mentioned something about Installer Script working this way for a reason (though I was unable to find that reason in closed PRs and issues) this is a WIP draft untill me or someone else test this with other scripts that use `postInstall`.
### What works
`Heroes of Might & Magic IV` no longer crash due to non-existing `.executable()`
### What was not tested
Other scripts with `postInstall`
### Test
- Operating system (and linux kernel version): Ubuntu 19.10
- Hardware (GPU/CPU): GTX 1080 ti, i7-7700K

### Ready for review

### DIFF
--- a/Engines/Wine/QuickScript/Installer Script/script.js
+++ b/Engines/Wine/QuickScript/Installer Script/script.js
@@ -95,9 +95,9 @@ module.default = class InstallerScript extends QuickScript {
             );
         }
 
-        this._createShortcut(wine.prefix());
-
         this._postInstall(wine, setupWizard);
+
+        this._createShortcut(wine.prefix());
 
         // back to generic wait (might have been changed in postInstall)
         setupWizard.wait(tr("Please wait..."));


plata:
@qparis can you remember if there was a reason to use this order?

madoar:
I don't see a reason why we should not be able to change the order. We should only face an issue with this if inside `postInstall` the created shortcut is accessed or the internal values of `QuickScript` are changed, which should not be the case anyway.

plata:
For me it's fine to try it. Could you take a look at the `postInstall` calls to see if there's something which might get broken?

madoar:
This is interesting. It seems like some installers use a different order:

https://github.com/PhoenicisOrg/scripts/blob/5fe4e1f8a9e0f1003e1e9c523c4afb61d15042f7/Engines/Wine/QuickScript/Origin%20Script/script.js#L48-L65

madoar:
ZipScript as well

plata:
Ok, then let's merge this.

PR #1180, opened 2020-01-05
ESLint: do not use var, prefer const

plata:


madoar:
Does this option detect whether a value is changed later on? (i.e. whether `let` should be used?)

plata:
The docs say:

> This rule is aimed at flagging variables that are declared using let keyword, but never reassigned after the initial assignment.

So I hope yes.

plata:
Actually, if you search for `let`, you will find that the PR contains it. So I think it should work.

PR #1182, opened 2020-01-05
Fix ESLint (include issues)

plata:


Zemogiter:
We can't remove that include from VK9 because the script copies files from tmp directory to system directory!

plata:
It's duplicate. The same line existed twice.

Zemogiter:
Whops you're correct

PR #1190, opened 2020-01-05
Add a Stage to Travis to execute ESLint

madoar:


plata:
Normally Codacy should already do this.

madoar:
If it does why then open https://github.com/PhoenicisOrg/scripts/pull/1169?

plata:
That was to test the GitHub action. The GitHub action was supposed to annotate the code directly in the PR. I hoped that this would be easier for committers compared to Codacy. However, it didn't work as expected because all GitHub actions I've tried required that the repository is an npm/yarn project.

madoar:
> required that the repository is an npm/yarn project.

This is not such a hard change.
Simply provide a `package.json` file with the bare minimum content. It would actually be useful to have such a file with a script to execute eslint and json-align

plata:
I understand that it's technically possible but does it make sense logically? Wouldn't you expect that you could build an npm module of the scripts repository?

madoar:
I think this depends on the documentation. The scripts don't need a build process so I wouldn't expect npm to be used for building purposes.

plata:
Could you provide a PR containing such a `package.json` so that we get a better basis for discussion?

madoar:
From a certain view point this is an interesting solution because it doesn't have the problem with the "possible issues" that for example Codacy has

PR #1192, opened 2020-01-05
Add package.json file to execute json-align and eslint

madoar:
This PR allows to execute `eslint` and `json-align` through the following commands:

```bash
npm install

npm run lint
npm run json-align
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name"        : "phoenicis-scripts",
     "version"     : "1.0.0",
-    "description" : "A collection of Scripts for Phoenicis",
+    "description" : "A collection of scripts for Phoenicis",
     "directories" : {
         "doc" : "docs"
     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -0,0 +1,29 @@
+{
+    "name"        : "phoenicis-scripts",
+    "version"     : "1.0.0",
+    "description" : "A collection of Scripts for Phoenicis",
```
plata: scripts should be lower case

madoar: Done

plata:
Should we update the documentation to use `npm run` instead of `eslint` and `json-align` directly?

madoar:
I think it would be a good idea. Especially because the you don't need to pass any parameters

plata:
Can you do it?

plata:
Please wait until we checked the GitHub action before merging this.

PR #1193, opened 2020-01-05
Use "ESlint Runner" GitHub action

plata:
see https://github.com/marketplace/actions/eslint-runner-action

I hope that this will make it easier for unexperienced users to understand why Codacy fails and adhere to our coding rules

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,20 @@
+name: Lint
+
+on: pull_request
+
+jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci --no-audit --prefer-offline
+      - uses: tinovyatkin/action-eslint@v1
+        with:
+          repo-token: ${{secrets.GITHUB_TOKEN}}
+          check-name: eslint # this is the check name from above 👆 where to post annotations
+          
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .DS_Store
 .idea
 *.iml
+
+node_modules/
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 *.iml
 
 node_modules/
-package-lock.json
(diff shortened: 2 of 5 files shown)


plata:
@madoar do you think this will work with #1192?

madoar:
I don't know.

I think it is fine to contain both in the same PR. Do you want to add my changes from #1192 to this PR?

plata:
I'm wondering because of
```
npm ci --no-audit --prefer-offline
```
Let's keep the changes separated. Nevertheless, if you add the package.json here as well, we can see if it will work (and then merge this PR after the other one).

madoar:
I never did any changes to a PR of someone else. Do you know how I can push to their branch?

plata:
It's just a branch in this repo (not in my fork). Shouldn't be a problem to edit it.

plata:
Can we fix the error?

> npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.

madoar:
It seems to work

plata:
But it doesn't show the annotations in #1169 like I would expect.

madoar:
Is it using the correct eslint configuration file?

plata:
I don't know. It's detecting errors but it doesn't show them somehow.

plata:
Or it's an error in the node execution?

madoar:
This is hard to tell. The issue is that I'm not sure where the given output:

```
##[error]Validation Failed
##[error]Node run failed with exit code 1
```

is coming from. It could be either eslint or the action. Additionally I don't know what arguments eslint is run with

plata:
I don't know. It's here: https://github.com/tinovyatkin/action-eslint/tree/master/src

madoar:
My best bet is that either we use `eslint` somehow unconventional or the developer of the action. I think the action fails at 

https://github.com/tinovyatkin/action-eslint/blob/07dd0b93c96ec0d081d6be631230729df668b061/src/main.ts#L105 

but I'm not 100\% sure

plata:
@tinovyatkin any idea what's happening here?

madoar:
@plata how should we continue here? Do we want to to continue with the Codacy comments or should we add another service?

plata:
Let's go for Codacy.

PR #1194, opened 2020-01-05
ESLint action (Run ESLint) test: break Wine script

plata:


plata:
This finds the errors but doesn't annotate which does not help.

madoar:
@plata can this PR be closed?

PR #1201, opened 2020-01-27
Fix Epic Games Launcher

### LINKED ISSUE #1200: Epic Games Launcher not working
Even after removing the checksum and installing corefonts and using "win7" or "win81", the installer does not run. When I try with plain Wine, the installer starts.

plata:
fixes #1200

### Description
* fix download URL
* use stable Wine
* install corefonts (not sure if actually required but listed on Wine AppDB)
* do not run OpenGL version (doesn't seem to be required anymore)

### What works
* Online installer

### What was not tested
* Local installer


Zemogiter:
Shouldn't wine use win7 by default since wine 4.0?

plata:
Could be. I will try without it.

PR #1205, opened 2020-03-15
Fix Remove Mono Verb for 32bit Wine prefixes

plata:


madoar:
Does this change really fix an issue or is this some cosmetic change?

plata:
It fixes an issue. You get an error with 32bit, see:
https://github.com/PhoenicisOrg/scripts/blob/d286456cdba8e0c6ab5aca56dd61e9aa6b355b12/Engines/Wine/Engine/Object/script.js#L342

madoar:
This makes no sense in my opinion. I mean you only move the definition of `system64directory` a few lines below. This should make no difference

plata:
Now, I call the function only if we are 64bit.

madoar:
Still it should make no difference. The value is used only in one place:

https://github.com/PhoenicisOrg/scripts/blob/359a6d16e7b947e59ede1de1e6dd44537132373a/Engines/Wine/Verbs/Remove%20Mono/script.js#L37

For this it should make absolutely no difference where the variable is defined.

plata:
It makes a big difference. On 32bit, the variable is not defined anymore and the error is not thrown. Feel free to try it yourself.

PR #1232, opened 2021-01-31
Add GitHub action to lint .js and .json

plata:
keep ESLint reviewdog for issues which cannot be fixed automatically


madoar (approved):
Autofix, nice!

Issue #1233, opened 2021-01-31
GitHub actions JSDoc update


plata:
We cannot push to master from GitHub actions. If we change the JSDoc in every PR, this will produce conflicts.

plata:
Maybe something like in https://github.com/cycjimmy/semantic-release-action#basic-usage

madoar:
Alternatively we could think about pushing to a separate `jsdoc` branch. In addition this should only activate on pushes to the master branch. Ideally a new PR would be created as well.

plata:
Yes, but this also has some drawbacks. E.g. what happens if the jsdoc branch is not merged before the action wants to create it again? In general, I would like to avoid manual interaction.

madoar:
> E.g. what happens if the jsdoc branch is not merged before the action wants to create it again? In general, I would like to avoid manual interaction.

In this case it should not recreate the `jsdoc` branch instead it should try to update it. If it is needed we could also think about doing a force push to the `jsdoc` branch...

PR #1235, opened 2021-01-31
Add GitHub action to update JSDoc

plata:
As explained in #1233, this should not be the final solution. However, it will work for the meantime while not too many PRs must be handled in parallel.


madoar:
Shouldn't the `docs` folder be excluded?

jsdoc_conf.json
```diff
@@ -6,6 +6,9 @@
     "plugins"      : [],
     "recurseDepth" : 10,
     "source"       : {
+        "exclude" : [
+            "docs"
```
plata, line 10: It is excluded.

madoar: But why does the PR contain so many changes to files located in that directory?

plata: Because the jsdoc got updated. It's pushed to that directory and then shown in GitHub pages.

madoar: Oh I see it is the output of the job... Somehow I mixed that up

madoar (approved):

PR #1248, opened 2021-03-29
Add GitHub action to compress images

plata:
Ensure that app miniatures are small to improve repo sync times.

### DIFF
--- a/.github/workflows/compress-images.yml
+++ b/.github/workflows/compress-images.yml
@@ -1,0 +1,21 @@
+name: Compress images
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


madoar:
Will this compress images only once or will this repeat with every push?
Should we compress all images or only ones that do not fulfill certain criteria? 
For example a certain file size threshold

plata:
Reading https://github.com/marketplace/actions/image-actions#compress-on-demand-or-on-schedule, my understanding is that it should not touch files which are already compressed enough. Not completely sure though. Maybe we also need `compressOnly: true`.

